PR #1, opened 2018-10-25
Bump ring to 0.13

csssuf:
Also bump untrusted to 0.6, as ring depends on it.

~_ring_ versions less than 0.12 do not support multiple version of _ring_ in a crate's dependency graph due to how its native crate works. _ring_ 0.13 fixes this, and bumping to this version will allow this crate to interoperate with other crates with _ring_ in their dependency graph more easily.~

Ah, looks like I misread and support for multiple versions hasn't landed in _ring_ yet, so this PR isn't especially useful on its own.


ramosbugs:
May as well bump the versions anyway. Thanks for the PR!

PR #8, opened 2020-03-04
Avoid possible duplicate version-conflicting dependencies

kaj:
Disable default features in the oath2 dependency.  This gets rid of the reqwest 0.9 dependency when using reqwest 0.10, getting rid of loads of transitive dependencies and some confusion.

Also remove direct http dependency, to avoid version confusion.  Use the http exported from oauth2 (or from reqwest, in a test) rather than having a direct http dependency.


kaj:
Oh, and hi!  Thanks for a promising crate!  :-)

ramosbugs:
Please see comments on ramosbugs/oauth2-rs#83. This crate aims to be a functional superset of the `oauth2` crate, and it should provide the same set of feature flags.

PR #9, opened 2020-03-13
Respect potential postfixes to conten types in user_info

HannesSommer:
## Problem
Getting a user_info fails on OIDC providers such as Gitlab that have an extended content type (e.g. with `charset`) fails with the exiting code because the content type is tested for equality.

## Implemented fix
I've simply moved existing code testing a header value for a given prefix into a function and used it in two relevant cases that were relying on equality before.


ramosbugs:
Thanks for the PR!

PR #10, opened 2020-03-18
Move retry to dev-depedencies.

kaj:
It is only used in test code, so not needed in the library itself.


kaj:
What?  How can moving a dependency to dev-dependencies decrease test coverage?  I'll have to investigate ...

ramosbugs:
Thanks! I wouldn't worry about the test coverage, though that is very odd. 🤷‍♂

PR #14, opened 2020-05-01
Adding exchange_password to Client

mettke:
This PR adds the `exchange_password` functionality from oauth2.

While this does not allow fetching an [id_token](https://docs.rs/openidconnect/1.0.0-alpha.13/openidconnect/trait.TokenResponse.html) (at least not using keycloak), it is possible to use the access_token with the [user_info](https://docs.rs/openidconnect/1.0.0-alpha.13/openidconnect/struct.Client.html#method.user_info) function.

Is there a reason why this function is not included at the moment?


ramosbugs:
thanks! seems like it was an oversight on my part

PR #16, opened 2020-05-23
Switching from Rc to Arc

mettke:
Hey man - got another patch for ya.

Tried to use the UserInfo endpoint using the async api endpoints (nice work btw) and it does not work as neither the `Rc` nor the inner function implements `Send` and `Sync`. This patch switches to `Arc` and explicitly requires `Send` and `Sync` for the inner function to make sure that it can be shared between threads.

Furthermore I had a look at the whole lib. This seems to be the only file where `Rc` is being used.


ramosbugs:
thanks!

PR #20, opened 2020-06-06
Updating itertools to 0.9

mettke:
Itertools:
https://github.com/rust-itertools/itertools/compare/0.7.11...v0.9.0
A lot of changes, deprecations and switch to 2018. Nothing that should affect this crate thou

`itertools::sorted` now returns an `IntoIter` instead of a `Vec`. Thus `iter` must not be called and we have to `collect` inside of the tests

### DIFF
--- a/src/registration.rs
+++ b/src/registration.rs
@@ -1030,7 +1030,8 @@ mod tests {
             ]
         );
         assert_eq!(
-            sorted(client_metadata.client_name().unwrap().clone()),
+            sorted(client_metadata.client_name().unwrap().clone())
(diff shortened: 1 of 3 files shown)


src/registration.rs
```diff
@@ -1030,7 +1030,8 @@ mod tests {
             ]
         );
         assert_eq!(
-            sorted(client_metadata.client_name().unwrap().clone()),
+            sorted(client_metadata.client_name().unwrap().clone())
```
ramosbugs, line 1033: oh nice! I'm glad they changed `sorted` to return an iterator

PR #21, opened 2020-06-07
Updating base64 to 0.12

mettke:
I had to split them. I underestimated the changes made to base64.

base64 removed the whitespace stripping. However the
auther provides a code snippet on the RELEASE Page
which fixes that [1]
`filter(|b| !b" \n\t\r\x0b\x0c".contains(b)`.

Unfortunatly that introduces an additional allocation
which, however, also exited before but was inside
base64.

The change to the test is necessary as the error
detection was improved. Now it is necessary that the
invalid signature is made out of valid base64

### DIFF
--- a/src/core/jwk.rs
+++ b/src/core/jwk.rs
@@ -263,12 +263,36 @@ impl CoreRsaPrivateSigningKey {
             return Err(format!("RSA private key must end with {}", RSA_FOOTER));
         }
         let base64_pem = &trimmed_pem[RSA_HEADER.len()..trimmed_pem.len() - RSA_FOOTER.len()];
-        let der = base64::decode_config(base64_pem, base64::MIME)
+        let base64_pem = Self::lax_base64_parsing(base64_pem);
+        let config = base64::STANDARD.decode_allow_trailing_bits(true);
+        let der = base64::decode_config(base64_pem, config)
             .map_err(|_| "Failed to decode RSA private key body as base64".to_string())?;
 
         let key_pair = ring_signature::RsaKeyPair::from_der(Input::from(&der))
             .map_err(|err| err.description_().to_string())?;
         Ok(Self { key_pair, rng, kid })
+    }
+
+    /// Filters characters from the base64 input string.
+    /// Charcters are specified according to lax base64 parsing.
+    ///
+    /// RFC 7468 Lax Parsing
+    fn lax_base64_parsing(input: &str) -> String {
+        input
+            .chars()
+            .filter(Self::keep_char_in_lax_base64_parsing)
+            .collect()
+    }
+
+    /// Returns whether a character is part of the base64 or should
+    /// be removed in accordance to lax base64 parsing.
+    ///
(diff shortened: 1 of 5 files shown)


mettke:
Ok I had a look at the test failures and they only happen sometimes as rp sometimes returns base64 which does not conform to canonical encoding that requires flexible parsing to accept. 

I did create an issue upstream https://github.com/openid-certification/oidctest/issues/227 and we may either wait for that to get in or use [Config::decode_allow_trailing_bits](https://docs.rs/base64/0.12.1/base64/struct.Config.html#method.decode_allow_trailing_bits) to get rid of this error and make the base64 decoding a bit more forgiving (as it currently is)

src/core/jwk.rs
```diff
@@ -263,7 +263,8 @@ impl CoreRsaPrivateSigningKey {
             return Err(format!("RSA private key must end with {}", RSA_FOOTER));
         }
         let base64_pem = &trimmed_pem[RSA_HEADER.len()..trimmed_pem.len() - RSA_FOOTER.len()];
-        let der = base64::decode_config(base64_pem, base64::MIME)
+        let base64_pem: String = base64_pem.chars().filter(|b| ![' ', '\n', '\t', '\r', '\x0b', '\x0c'].contains(b)).collect();
```
ramosbugs, line 266: it looks like https://tools.ietf.org/html/rfc7468 is the most relevant standard. [Figure 2 (Lax)](https://tools.ietf.org/html/rfc7468#section-3) is probably what we want here, and I think that's consistent with this code. mind adding a comment that this is RFC 7468 "Lax" parsing?

I'm curious about the relative performance of `[...].contains()` vs. a pre-compiled regex vs. Rust pattern matching, but it probably doesn't matter too much since this code doesn't seem like it would be on the critical path for most client applications.

src/jwt.rs
```diff
@@ -711,7 +711,7 @@ pub mod tests {
             .chars()
             .take(TEST_JWT.len() - 1)
             .collect::<String>()
-            + "f";
+            + "w";
```
ramosbugs, line 714: I'm curious why `f` doesn't adequately corrupt the signature

mettke: I just wrote a small section about that on https://github.com/openid-certification/oidctest/issues/227 so if you want the exact reasoning behind it you may want to read it there as I'm just going for a short version here:

Of course it does adequately corrupt the signature. The problem is that it never arrives at the signature checking as the base64 is invalid thus the decoding fails. Its the same problem that we have on the rs endpoint but there I was able to fix it by replacing it with valid base64. I will reset it once we switch to forgiving parsing

src/core/jwk.rs
mettke: I added the comment and switched to pattern matching. Guess from those three pattern matching is the fastest.

mettke:
I added `decode_allow_trailing_bits` only to the methods for decoding. The encoding methods don't bother with it anyways

ramosbugs:
great work, thanks!

PR #26, opened 2020-07-07
Replace failure with thiserror

Weasy666:
This PR is based on @mettke's and finalizes the move to oauth4 with removing `failure` as a dependency and replacing it with `thiserror` and `anyhow`(in the docs).

To make this work, i had to change `oauth2` from version 3 to the github `main`-branch and as result of this i also had to remove the `futures-03`-feature. I guess you want to wait with merging this, until you have released an `alpha`.version of `oauth2` to crates.io. Just let me know how i should proceed from here on.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,8 +23,7 @@ nightly = []
 [dependencies]
 base64 = "0.12"
 chrono = "0.4"
-failure = "0.1"
-failure_derive = "0.1"
+thiserror = "1.0"
 http = "0.2"
 itertools = "0.9"
 log = "0.4"
@@ -44,3 +43,4 @@ env_logger = "0.7"
 pretty_assertions = "0.6"
 reqwest_ = { package = "reqwest", features = ["blocking", "rustls-tls"], version = "0.10", default-features = false }
 retry = "1.0"
+anyhow = "1.0"
--- a/src/discovery.rs
+++ b/src/discovery.rs
@@ -317,7 +317,7 @@ where
     ) -> Result<Self, DiscoveryError<RE>>
     where
         HC: Fn(HttpRequest) -> Result<HttpResponse, RE>,
-        RE: std::error::Error + Send + Sync + 'static,
+        RE: std::error::Error + 'static,
     {
         let discovery_url = issuer_url
             .join(CONFIG_URL_SUFFIX)
@@ -345,7 +345,7 @@ where
     where
         F: Future<Output = Result<HttpResponse, RE>>,
         HC: Fn(HttpRequest) -> F + 'static,
-        RE: std::error::Error + Send + Sync + 'static,
+        RE: std::error::Error + 'static,
     {
         let discovery_url = issuer_url
             .join(CONFIG_URL_SUFFIX)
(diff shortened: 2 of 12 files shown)


Cargo.toml
```diff
@@ -15,25 +15,19 @@ all-features = true
...
-failure_derive = "0.1"
-futures-0-1 = { version = "0.1", optional = true, package = "futures" }
-futures-0-3 = { version = "0.3", optional = true, package = "futures" }
-http = "0.1"
+thiserror = "1.0"
+http = "0.2"
 itertools = "0.9"
 log = "0.4"
-oauth2 = "3.0"
+oauth2 = { git = "https://github.com/ramosbugs/oauth2-rs", branch = "main", version = "4.0.0-alpha.1" }
```
ramosbugs, line 30:
```suggestion
oauth2 = "=4.0.0-alpha.1"
```

src/discovery.rs
```diff
@@ -320,7 +318,7 @@ where
     ) -> Result<Self, DiscoveryError<RE>>
     where
         HC: Fn(HttpRequest) -> Result<HttpResponse, RE>,
-        RE: Fail,
+        RE: std::error::Error + Send + Sync + 'static,
```
ramosbugs, line 321: I think we can get rid of these `Send` and `Sync` bounds now after https://github.com/ramosbugs/oauth2-rs/pull/109

Weasy666: Check! Should i also remove the 3 `Send + Sync`-bounds in `verification.rs` on following lines?https://github.com/ramosbugs/openidconnect-rs/blob/616133adb6f21a66b1aa9ff0dc4641c5beda3ed4/src/verification.rs#L134
https://github.com/ramosbugs/openidconnect-rs/blob/616133adb6f21a66b1aa9ff0dc4641c5beda3ed4/src/verification.rs#L199
https://github.com/ramosbugs/openidconnect-rs/blob/616133adb6f21a66b1aa9ff0dc4641c5beda3ed4/src/verification.rs#L699

Weasy666:
I just noticed, that i missed a few `Fail`s, which are behind the `futures-03`-feature flag. I also missed to remove the `cfg`-flags in the source files 🤦‍♂️. Sorry about that.
So...i would like to know how i should go about this. I could just remove all occurrences of `#[cfg(feature = "futures-03")]` and call it a day (i actually did this locally and everything works as expected), but then `blocking` and `async` are always on.

mettke:
@Weasy666 The flag does not exist on my PR anymore. So the best would be not to deal with it but it rebase mine.

Weasy666:
Nice! Did you introduce a new flag, or are both always on?

mettke:
Both are always on. Same as in oauth2-rs. If someone would like to get rid of blocking than a custom implementation is always possible. Also the future crate got removed as the future links to the one in the std library, which can be used instead.

Weasy666:
Ok, that is what i also had tested locally, so there shouldn't be much things to change on my part. Will rebase it now.

ramosbugs:
looks like quite a few merge conflicts now :-(

Weasy666:
Hm...well...damn! I will take a look at it tomorrow and maybe just rebase it again.

Weasy666:
@ramosbugs Done! Was a painless rebase. Looks like Git just had some kind of hiccup ¯\\_(ツ)\_/¯

ramosbugs:
thanks!

PR #24, opened 2020-06-23
Migrating to new oauth4 version

mettke:
(Build failure expected)

This time separating is a bit difficult so its all combined into one. 

* Removed future-01
* Removed reqwest-09
* Updated http to 0.2
* Updated dev dependencies to latest version

When you release oauth2 and switch openidconnect-rs to version 4 remember to set `default-features = false` ;)

PR #29, opened 2020-08-11
Require Sync + Send for sharing verifier between threads

tsing:
Hi,

I would like to share the token verifier across threads and have to add Sync + Send for these inner functions.

I'm not sure if the change is acceptable, and I noticed that #16 added some similar changes on this, but was removed from #24

### DIFF
--- a/src/verification.rs
+++ b/src/verification.rs
@@ -131,7 +131,7 @@ where
     iss_required: bool,
     issuer: IssuerUrl,
     is_signature_check_enabled: bool,
-    other_aud_verifier_fn: Arc<dyn Fn(&Audience) -> bool + 'a>,
+    other_aud_verifier_fn: Arc<dyn Fn(&Audience) -> bool + 'a + Send + Sync>,
     signature_keys: JsonWebKeySet<JS, JT, JU, K>,
 }
 impl<'a, JS, JT, JU, K> JwtClaimsVerifier<'a, JS, JT, JU, K>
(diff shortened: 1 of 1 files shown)


ramosbugs:
thanks for the PR! I think this is technically a breaking change, but I don't foresee it causing issues (plus we're still in the 4.x alpha release series, which allows for breaking changes).

mind fixing the build breakage this PR leads to? https://travis-ci.org/github/ramosbugs/openidconnect-rs/jobs/716863272#L738

once CI passes, I'll be happy to merge it :-)

ramosbugs:
@mettke do you happen to remember why we removed `Send + Sync` from the verifier functions in #24? it seems like that may have been done as part of porting ramosbugs/oauth2-rs#109 to this crate, but it seems like we probably only needed to remove those trait bounds for the async HTTP implementations, not for the synchronous functions that verify audience, expiration time, etc.

I think the issue @tsing is running into is that if the client code needs to have an `Arc<dyn Fn(&Audience) -> bool + 'a + Send + Sync>`, there's no way to cast that to an `Arc<dyn Fn(&Audience) -> bool + 'a>` before passing it to the library... or is there?

mettke:
Yeah removing those Send + Sync most likely was an oversight. While removing most of the Send + Sync was ok, I should have left the functions as they are as they require Send + Sync :sweat_smile:. Merging sounds good to me

tsing:
@ramosbugs @mettke thank you for looking into this, now the tests passed.

ramosbugs:
thanks!

PR #30, opened 2020-08-30
Client credentials grant

GrandChaman:
Hello ! First of all, thank you very much for that nice crate you made !
I just thought the `exchange_client_credentials` was missing from the `openidconnect::Client`.


ramosbugs:
thanks for the PR! this does seem like an oversight.

PR #32, opened 2020-10-21
Add EC signature verification

ubamrein:
This PullRequest adds the possibility to verify JWTs signed with ecdsa.

### DIFF
--- a/src/core/crypto.rs
+++ b/src/core/crypto.rs
@@ -91,7 +91,7 @@ pub fn verify_ec_signature(
 ) -> Result<(), SignatureVerificationError> {
     let (x, y, crv) = ec_public_key(&key).map_err(SignatureVerificationError::InvalidKey)?;
     if *crv == CoreJsonCurveType::P521{
-        return Err(SignatureVerificationError::UnsupportedAlg("Only P256 and P384 are supported for now".to_string()));
+        return Err(SignatureVerificationError::UnsupportedAlg("P521".to_string()));
     }
     let mut pk = vec![0x04];
     pk.extend(x.deref());
--- a/src/core/crypto.rs
+++ b/src/core/crypto.rs
@@ -101,8 +101,8 @@ pub fn verify_ec_signature(
     signature: &[u8],
 ) -> Result<(), SignatureVerificationError> {
     let (x, y, crv) = ec_public_key(&key).map_err(SignatureVerificationError::InvalidKey)?;
-    if *crv != CoreJsonCurveType::P256 {
-        return Err(SignatureVerificationError::UnsupportedAlg("Only P256 is supported for now".to_string()));
+    if *crv == CoreJsonCurveType::P521{
+        return Err(SignatureVerificationError::UnsupportedAlg("Only P256 and P384 are supported for now".to_string()));
     }
     let mut pk = vec![0x04];
     pk.extend(x.deref());
--- a/src/core/crypto.rs
+++ b/src/core/crypto.rs
@@ -96,6 +96,7 @@ pub fn verify_rsa_signature(
 
 pub fn verify_ec_signature(
     key: &CoreJsonWebKey,
+    params: &'static ring_signature::EcdsaVerificationAlgorithm,
     msg: &[u8],
     signature: &[u8],
 ) -> Result<(), SignatureVerificationError> {
@@ -106,7 +107,7 @@ pub fn verify_ec_signature(
     let mut pk = vec![0x04];
     pk.extend(x.deref());
     pk.extend(y.deref());
-    let public_key = ring_signature::UnparsedPublicKey::new(&ring_signature::ECDSA_P256_SHA256_FIXED, pk);
+    let public_key = ring_signature::UnparsedPublicKey::new(params, pk);
     public_key.verify(msg, signature)
     .map_err(|_| SignatureVerificationError::CryptoError("EC Signature was wrong".to_string()))
 }
--- a/src/core/crypto.rs
+++ b/src/core/crypto.rs
@@ -5,7 +5,7 @@ use ring::signature as ring_signature;
 use crate::types::Base64UrlEncodedBytes;
 use crate::{JsonWebKey, SignatureVerificationError, SigningError};
 
-use super::{CoreJsonWebKey, CoreJsonWebKeyType};
+use super::{jwk::CoreJsonCurveType, CoreJsonWebKey, CoreJsonWebKeyType};
 
 use std::ops::Deref;
 
@@ -57,6 +57,26 @@ fn rsa_public_key(
     }
 }
 
+fn ec_public_key(
+    key: &CoreJsonWebKey,
+) -> Result<(&Base64UrlEncodedBytes, &Base64UrlEncodedBytes, &CoreJsonCurveType), String> {
+    if *key.key_type() != CoreJsonWebKeyType::EllipticCurve {
+        Err("EC key required".to_string())
+    } else if let Some(x) = key.x.as_ref() {
+        if let Some(y) = key.y.as_ref() {
+            if let Some(crv) = key.crv.as_ref() {
+                Ok((x,y, crv))
+            } else {
+                Err("CurveType is missing".to_string())
+            }
+        } else {
+            Err("EC `x` part is missing".to_string())
+        }
+    } else {
+        Err("EC `y` part is missing".to_string())
+    }
+}
+
 pub fn verify_rsa_signature(
     key: &CoreJsonWebKey,
     params: &ring_signature::RsaParameters,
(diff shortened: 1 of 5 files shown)


src/core/jwk.rs
```diff
@@ -49,6 +50,31 @@ pub struct CoreJsonWebKey {
     )]
     pub(crate) e: Option<Base64UrlEncodedBytes>,
 
+    //Elliptic Curve
+    #[serde(
+        skip_serializing_if = "Option::is_none"
```
ramosbugs, line 55: let's add `default` and `deserialize_with = "deserialize_option_or_none"` here too, I think. this struct currently uses `deserialize_option_or_none` for all of the cipher-specific fields for the reason described on lines 36-39.

src/core/crypto.rs
```diff
@@ -61,6 +61,26 @@ fn rsa_public_key(
...
+        Err("EC key required".to_string())
+    } else if let Some(x) = key.x.as_ref() {
+        if let Some(y) = key.y.as_ref() {
+            if let Some(crv) = key.crv.as_ref() {
+                Ok((x,y, crv))
+            } else {
+                Err("CurveType is missing".to_string())
+            }
+        } else {
+            Err("EC `x` part is missing".to_string())
```
ramosbugs, line 77: I think this error should say `y`. the code above is structured similarly, but we'd probably avoid this sort of bug by expressing it as:
```
let x = key.x.as_ref().ok_or_else(|| Err("EC `x` part is missing".to_string()))?;
let y = key.y.as_ref().ok_or_else(|| Err("EC `y` part is missing".to_string()))?;
let crv = key.crv.as_ref().ok_or_else(|| Err("EC `crv` part is missing".to_string()))?;
Ok((x, y, crv))
```

src/core/crypto.rs
```diff
@@ -61,6 +61,26 @@ fn rsa_public_key(
...
+        }
+    } else {
+        Err("EC `y` part is missing".to_string())
```
ramosbugs, line 80: `x` here

src/core/crypto.rs
```diff
@@ -74,3 +94,21 @@ pub fn verify_rsa_signature(
...
+
+pub fn verify_ec_signature(
+    key: &CoreJsonWebKey,
+    params: &'static ring_signature::EcdsaVerificationAlgorithm,
+    msg: &[u8],
+    signature: &[u8],
+) -> Result<(), SignatureVerificationError> {
+    let (x, y, crv) = ec_public_key(&key).map_err(SignatureVerificationError::InvalidKey)?;
+    if *crv == CoreJsonCurveType::P521{
+        return Err(SignatureVerificationError::UnsupportedAlg("Only P256 and P384 are supported for now".to_string()));
```
ramosbugs, line 106: I would just make this string `"P521"` since the `Display` output will turn that into `Unsupported signature algorithm: P521` (which is consistent with other instances of this error)

src/core/crypto.rs
```diff
@@ -74,3 +94,21 @@ pub fn verify_rsa_signature(
...
+    }
+    let mut pk = vec![0x04];
```
ramosbugs, line 108: mind adding a comment linking to the relevant RFC/doc that defines this key param concatenation?

ramosbugs (commented):
thanks for the PR!

before merging this functionality, I think we'll need some comprehensive tests. ideally, that would consist of ECDSA equivalents of the following tests:
- `test_core_jwk_deserialization_rsa`
- `test_rsa_pkcs1_verification` (see https://tools.ietf.org/html/rfc7520#section-4.3 for some potentially useful test inputs, though those appear to use P-521)

ubamrein:
Pushed the suggested changes :) 

I will add some tests later on!

ubamrein:
Since I did not find an existing signature example on the RFC (since P521 is not supported) I just generated some random signatures using the `pcks1_signing_input` and the `ring` library. 

If you are not ok with it, I can generate signatures with Java or so, to test the verification independently of `ring` (since now we are basically just testing the ring lib). 

Anyhow, here the code I used to generate the signatures:
```rust
use ring::signature::{EcdsaKeyPair, ECDSA_P256_SHA256_FIXED_SIGNING, ECDSA_P384_SHA384_FIXED_SIGNING, KeyPair};
use ring::rand;
use simple_asn1::from_der;
use base64;

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let rng = rand::SystemRandom::new();
    let pkcs8_bytes = EcdsaKeyPair::generate_pkcs8(&ECDSA_P384_SHA384_FIXED_SIGNING, &rng).unwrap();
    
    let key_pair = EcdsaKeyPair::from_pkcs8(&ECDSA_P384_SHA384_FIXED_SIGNING, pkcs8_bytes.as_ref()).unwrap();
    let key_bytes = key_pair.public_key().as_ref();
    println!("PublicKey");
    let x = &key_bytes[1..=48];
    let y = &key_bytes[49..];
    println!("X: {:02x?}", x);
    println!("Y: {:02x?}", y);
   let jwk  = format!("{{
        \"kty\": \"EC\",
        \"kid\": \"bilbo.baggins@hobbiton.example\",
        \"use\": \"sig\",
        \"crv\" : \"P-384\",
        \"x\": \"{}\",
        \"y\": \"{}\"
    }}", base64::encode_config(x, base64::URL_SAFE_NO_PAD), base64::encode_config(y, base64::URL_SAFE_NO_PAD));
    println!("{}", jwk);

    let pkcs1_signing_input = "eyJhbGciOiJSUzI1NiIsImtpZCI6ImJpbGJvLmJhZ2dpbnNAaG9iYml0b24uZX\
                                   hhbXBsZSJ9.\
                                   SXTigJlzIGEgZGFuZ2Vyb3VzIGJ1c2luZXNzLCBGcm9kbywgZ29pbmcgb3V0IH\
                                   lvdXIgZG9vci4gWW91IHN0ZXAgb250byB0aGUgcm9hZCwgYW5kIGlmIHlvdSBk\
                                   b24ndCBrZWVwIHlvdXIgZmVldCwgdGhlcmXigJlzIG5vIGtub3dpbmcgd2hlcm\
                                   UgeW91IG1pZ2h0IGJlIHN3ZXB0IG9mZiB0by4";
    let sig = key_pair.sign(&rng, pkcs1_signing_input.as_ref()).unwrap();
    println!("Signature: {}", base64::encode_config(sig.as_ref(), base64::URL_SAFE_NO_PAD));
   
    Ok(())
}

```

ramosbugs (commented):
sorry for the delay! I've been heads-down in some other projects the past few weeks.

thanks for addressing my comments; the tests looks great. let's just address the new security issue I raised, and then this PR should be ready to merge!

I didn't notice before that this PR was on the `support/1.x` branch. since it doesn't introduce any breaking changes, I think it's fine to add this new functionality to 1.x. I'll need to forward port it to 2.x though, which could be annoying since the ring versions differ. I would definitely not object if you wanted to target this change to 2.x :-)

ramosbugs:
oh, please merge and/or rebase your branch too to pull in the Travis config changes I just made, which will run `rustfmt` and `clippy` as we're doing on the `main` branch

src/core/crypto.rs
```diff
@@ -74,3 +83,27 @@ pub fn verify_rsa_signature(
...
+/// is used (Section 2.3.4 of https://www.secg.org/sec1-v2.pdf).
+
+pub fn verify_ec_signature(
+    key: &CoreJsonWebKey,
+    params: &'static ring_signature::EcdsaVerificationAlgorithm,
+    msg: &[u8],
+    signature: &[u8],
+) -> Result<(), SignatureVerificationError> {
+    let (x, y, crv) = ec_public_key(&key).map_err(SignatureVerificationError::InvalidKey)?;
+    if *crv == CoreJsonCurveType::P521{
```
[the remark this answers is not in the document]
ubamrein: So your concern only applies to cases, where the curve used in the public key mismatches the curved used for the JWT, as the other cases are restricted by default (one has to exactly specify which algorithm to use) right?

The thing about elliptic curve cryptography is that you cannot at all use points from different curves. As the underlying math (and the first check before verification according to the ring documentation the [NIST](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar2.pdf) guidelines are used to verify the public-key) guarantees that the signature and the public key need to originate from the same curve. 

But, better safe than sorry :) I will check for the curve paramter found in the JWK and compare it to the one given as params.

ubamrein:
> I didn't notice before that this PR was on the `support/1.x` branch. since it doesn't introduce any breaking changes, I think it's fine to add this new functionality to 1.x. I'll need to forward port it to 2.x though, which could be annoying since the ring versions differ. I would definitely not object if you wanted to target this change to 2.x :-)

Since internally we are using it in production, I did not want to add a "alpha" version of the library :D on crates.io 1.0.1 is the "latest-stable" so I was just using that one. I can try to forward port it to 2.0 as well :)

ramosbugs:
thanks! I cherry-picked this into 2.x, and I'll cut some releases after CI passes. this feature will be in 1.1.0 and 2.0.0-alpha.2.

PR #33, opened 2020-11-18
Fixed the 'description_' method not available anymore in the 'ring' crate

GrandChaman:
The `ring` crate was updated, the `KeyRejected::description_` method was deleted. `openidconnect-rs` can't compile without this method.


ramosbugs:
thank you!

Issue #28, opened 2020-08-04
Better examples needed


frenchtoastbeer:
I eventually was able to take the samples of how to use this (specifically the await example) and turn it into something at compiled.  A few things were necessary in order to do that, one of which centered around error handling.  I'm sure I did it wrong, but without a better option I'm just going to posit here what I did so that you'll know that A - it didn't appear to me that your example usage code was functional and B - maybe having my sample code helps get to a better solution faster.

cargo.toml

```
[dependencies]
openidconnect = { version = "1.0", features = ["futures-03","reqwest-010"], default-features = false }
oauth2 = "3.0"
failure = "0.1"
```

auth.rs
```
pub async fn authenticate() -> Result<(oauth2::AccessToken, Option<oauth2::RefreshToken>), MyErr> {
    use openidconnect::{
        AccessTokenHash,
        AsyncCodeTokenRequest,
        AuthorizationCode,
        ClientId,
        ClientSecret,
        CsrfToken,
        Nonce,
        IssuerUrl,
        PkceCodeChallenge,
        RedirectUrl,
        Scope,
    };
    use openidconnect::core::{
      CoreAuthenticationFlow,
      CoreClient,
      CoreProviderMetadata,
    };
    use openidconnect::reqwest::async_http_client;
    
    // Use OpenID Connect Discovery to fetch the provider metadata.
    use openidconnect::{OAuth2TokenResponse, TokenResponse};
    let provider_metadata = CoreProviderMetadata::discover_async(
        IssuerUrl::new("https://accounts.example.com".to_string())?,
        async_http_client,
    )
    .await?;
    
    // Create an OpenID Connect client by specifying the client ID, client secret, authorization URL
    // and token URL.
    let client =
        CoreClient::from_provider_metadata(
            provider_metadata,
            ClientId::new("client_id".to_string()),
            Some(ClientSecret::new("client_secret".to_string())),
        )
        // Set the URL the user will be redirected to after the authorization process.
        .set_redirect_uri(RedirectUrl::new("http://redirect".to_string())?);
    
    // Generate a PKCE challenge.
    let (pkce_challenge, pkce_verifier) = PkceCodeChallenge::new_random_sha256();
    
    // Generate the full authorization URL.
    let (auth_url, csrf_token, nonce) = client
        .authorize_url(
            CoreAuthenticationFlow::AuthorizationCode,
            CsrfToken::new_random,
            Nonce::new_random,
        )
        // Set the desired scopes.
[...]

PR #36, opened 2021-01-30
Draft: Enhance example in crate documentation for requesting UserInfo

benjaminSchilling33:
@ramosbugs 
As promised in https://github.com/ramosbugs/openidconnect-rs/issues/28#issuecomment-770036819, I started extending the example with a UserInfo request.

How are you testing this example?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -190,6 +190,41 @@
 //!     claims.email().map(|email| email.as_str()).unwrap_or("<not provided>"),
 //! );
 //!
+//! // If available, we can use the UserInfo endpoint to request additional information
+//!
+//! // First we need a struct that contains the AdditionalClaims we want to retrieve
+//! #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+//! struct MyClaims {
+//!     roles: String,
+//! }
+//! // We use an empty implementation
+//! impl AdditionalClaims for MyClaims {}
+//!
+//! // Now we create a new user_info request using the client
+//! let userinfo_request = match client.user_info(
+//! AccessToken::new(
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -190,6 +190,41 @@
...
+//! impl AdditionalClaims for MyClaims {}
+//!
+//! // Now we create a new user_info request using the client
+//! let userinfo_request = match client.user_info(
+//! AccessToken::new(
+//!   // We re-use the access_token we used for requesting the IdToken
+//!   token_response.access_token().secret().to_string()),
+//!   None,
+//! ) {
+//!   Ok(o) => o,
```
ramosbugs, line 210: let's replace this match and the one below with the more common `.map_err(|err| ...)?` pattern

ramosbugs (commented):
thanks for the PR! the examples are compile-tested as part of `cargo test --doc`. please take a look at the related CI errors: https://travis-ci.org/github/ramosbugs/openidconnect-rs/jobs/756820151.

since this example is illustrating custom claims and not just the default user info request (with [`CoreUserInfoClaims`](https://docs.rs/openidconnect/2.0.0-alpha.2/openidconnect/core/type.CoreUserInfoClaims.html)), I think it should be separate from the "Getting Started" example. it would be fine to add a `CoreUserInfoClaims` example to Getting Started since it's a common enough need, but let's separate them if you'd like to illustrate how to implement an extension (also a useful example to have!)

src/lib.rs
benjaminSchilling33: How about now?

ramosbugs:
thanks! I merged a slightly simplified version in 6fb05a5153e1b53a7ac543e5db674bc6a9f64cbf

PR #39, opened 2021-02-15
Minimal support for exposing new oauth2-rs crate token introspection and revocation features (resolves #37).

### LINKED ISSUE #37: Support RFC 7009 OAuth 2.0 Token Revocation
Hi,

This is really a feature request for the oauth2 crate but I am a user of the openidconnect crate and thus want the feature to be reachable through the openidconnect crate, hence why I raise it here.

In my own crate I am implementing support for https://tools.ietf.org/html/rfc7009. Unfortunately, unless I am missing something, I am unable to leverage the basic functionality of making an authenticated request to the OAuth 2.0 / OpenID Connect provider that is defined in [`lib.rs`](https://github.com/ramosbugs/oauth2-rs/blob/32b417c23d967d3fb63d0995210a98803082608e/src/lib.rs#L1564) in the `endpoint_request()` function as it is a private function. I either have to fork the `oauth2` and `openidconnect` crates or, as I have done now, copy code from `lib.rs` into my own crate.

Would there be interest in my contributing code for OAuth 2.0 Token Revocation support to your crates?

Am I missing something and should handle this a different way? If not, should this be easier to extend in this way without adding to your crates?

Thanks,

Ximon

ximon18:
This is a minimal first effort to address #37 now that support for token revocation was merged into `oauth2-rs` release `4.0.0-alpha.4`.

`cargo fmt` has NOT been applied to the code as doing so causes:

```
        provider_metadata: ProviderMetadata<A, AD, CA, CN, CT, G, JE, JK, JS, JT, JU, K, RM, RT2, S>,
```

To become 10 or so lines with `A`, `AD` etc each on their own line which is ugly, and I didn't find a way to tell `rustfmt` to ignore just that one change (the `skip` support in `rustfmt` cannot be used for a single parameter of a function).

This PR lacks additional richer documentation, e.g. example code, in particular referencing the fact that you will need to customize discovery response handling if you wish to use the non-standard `revocation_endpoint` attribute of the discovery response as the URL to be used by `Client`.

I'm also unsure about error handling. In `oauth2-rs` if the revocation URL has not been set or is not HTTPS this was implemented such that it causes a `TokenRequestError`. However, I see that `fn user_info()` handles a similar case in a different way by returning a `NoUserInfoEndpoint` error.

I also didn't update the main docs to reflect the addition of token introspection and revocation support, as I did in the `oauth2-rs` crate.

I also realize that I haven't run or updatd the tests or attempted to build with anything except default features. I have however successfully used this modified version of the create with my own application.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ thiserror = "1.0"
 http = "0.2"
 itertools = "0.9"
 log = "0.4"
-oauth2 = { version = "=4.0.0-alpha.4", default-features = false }
+oauth2 = { version = "=4.0.0-alpha.5", default-features = false }
 rand = "0.7"
 ring = "0.16"
 serde = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "openidconnect"
-version = "2.0.0-alpha.3"
+version = "2.0.0-alpha.4"
 authors = ["David A. Ramos <ramos@cs.stanford.edu>"]
 description = "OpenID Connect library"
 license = "MIT"
@@ -27,7 +27,7 @@ thiserror = "1.0"
 http = "0.2"
 itertools = "0.9"
 log = "0.4"
-oauth2 = { version = "4.0.0-alpha.3", default-features = false }
+oauth2 = { version = "4.0.0-alpha.4", default-features = false }
 rand = "0.7"
 ring = "0.16"
 serde = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ thiserror = "1.0"
 http = "0.2"
 itertools = "0.9"
 log = "0.4"
-oauth2 = { version = "=4.0.0-alpha.5", default-features = false }
+oauth2 = { version = "=4.0.0-alpha.6", default-features = false }
 rand = "0.7"
 ring = "0.16"
 serde = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "openidconnect"
-version = "2.0.0-alpha.4"
+version = "2.0.0-alpha.3"
 authors = ["David A. Ramos <ramos@cs.stanford.edu>"]
 description = "OpenID Connect library"
 license = "MIT"
@@ -27,7 +27,7 @@ thiserror = "1.0"
 http = "0.2"
 itertools = "0.9"
 log = "0.4"
-oauth2 = { version = "4.0.0-alpha.4", default-features = false }
+oauth2 = { version = "=4.0.0-alpha.4", default-features = false }
 rand = "0.7"
 ring = "0.16"
 serde = "1.0"
--- a/examples/google.rs
+++ b/examples/google.rs
@@ -234,7 +234,7 @@ fn main() {
 
             client
                 .revoke_token(token_to_revoke)
-                .unwrap()
+                .expect("no revocation_uri configured")
                 .request(http_client)
                 .unwrap_or_else(|err| {
                     handle_error(&err, "Failed to contact token revocation endpoint");
--- a/examples/google.rs
+++ b/examples/google.rs
@@ -19,13 +19,20 @@ use std::io::{BufRead, BufReader, Write};
 use std::net::TcpListener;
 use std::process::exit;
 
+use serde::{Deserialize, Serialize};
 use url::Url;
 
-use openidconnect::core::{CoreClient, CoreIdTokenClaims, CoreIdTokenVerifier, CoreResponseType};
-use openidconnect::{reqwest::http_client, RevocationUrl};
+use openidconnect::core::{
+    CoreAuthDisplay, CoreClaimName, CoreClaimType, CoreClient, CoreClientAuthMethod, CoreGrantType,
+    CoreIdTokenClaims, CoreIdTokenVerifier, CoreJsonWebKey, CoreJsonWebKeyType, CoreJsonWebKeyUse,
+    CoreJweContentEncryptionAlgorithm, CoreJweKeyManagementAlgorithm, CoreJwsSigningAlgorithm,
+    CoreResponseMode, CoreResponseType, CoreRevocableToken, CoreSubjectIdentifierType,
+};
+use openidconnect::reqwest::http_client;
 use openidconnect::{
-    AuthenticationFlow, AuthorizationCode, ClientId, ClientSecret, CsrfToken, IssuerUrl, Nonce,
-    OAuth2TokenResponse, RedirectUrl, Scope,
+    AdditionalProviderMetadata, AuthenticationFlow, AuthorizationCode, ClientId, ClientSecret,
+    CsrfToken, IssuerUrl, Nonce, OAuth2TokenResponse, ProviderMetadata, RedirectUrl, RevocationUrl,
+    Scope,
 };
 
 fn handle_error<T: std::error::Error>(fail: &T, msg: &'static str) {
@@ -41,16 +48,6 @@ fn handle_error<T: std::error::Error>(fail: &T, msg: &'static str) {
 
 // Teach openidconnect-rs about the Google custom extension to the OpenID Discovery response
 // that we can use as the RFC 7009 OAuth 2.0 Token Revocation endpoint.
-use openidconnect::core::{
-    CoreAuthDisplay, CoreClaimName, CoreClaimType, CoreClientAuthMethod, CoreGrantType,
-    CoreJsonWebKey, CoreJsonWebKeyType, CoreJsonWebKeyUse, CoreJweContentEncryptionAlgorithm,
-    CoreJweKeyManagementAlgorithm, CoreJwsSigningAlgorithm, CoreResponseMode, CoreRevocableToken,
-    CoreSubjectIdentifierType,
-};
-use openidconnect::{AdditionalProviderMetadata, ProviderMetadata};
-
-use serde::{Deserialize, Serialize};
-
 #[derive(Clone, Debug, Deserialize, Serialize)]
 struct RevocationEndpointProviderMetadata {
     revocation_endpoint: String,
--- a/examples/google.rs
+++ b/examples/google.rs
@@ -21,10 +21,8 @@ use std::process::exit;
 
 use url::Url;
 
-use openidconnect::core::{
-    CoreClient, CoreIdTokenClaims, CoreIdTokenVerifier, CoreProviderMetadata, CoreResponseType,
-};
-use openidconnect::reqwest::http_client;
+use openidconnect::core::{CoreClient, CoreIdTokenClaims, CoreIdTokenVerifier, CoreResponseType};
+use openidconnect::{reqwest::http_client, RevocationUrl};
 use openidconnect::{
     AuthenticationFlow, AuthorizationCode, ClientId, ClientSecret, CsrfToken, IssuerUrl, Nonce,
     OAuth2TokenResponse, RedirectUrl, Scope,
@@ -41,6 +39,41 @@ fn handle_error<T: std::error::Error>(fail: &T, msg: &'static str) {
     exit(1);
 }
 
+// Teach openidconnect-rs about the Google custom extension to the OpenID Discovery response
+// that we can use as the RFC 7009 OAuth 2.0 Token Revocation endpoint.
+use openidconnect::core::{
+    CoreAuthDisplay, CoreClaimName, CoreClaimType, CoreClientAuthMethod, CoreGrantType,
+    CoreJsonWebKey, CoreJsonWebKeyType, CoreJsonWebKeyUse, CoreJweContentEncryptionAlgorithm,
+    CoreJweKeyManagementAlgorithm, CoreJwsSigningAlgorithm, CoreResponseMode, CoreRevocableToken,
+    CoreSubjectIdentifierType,
+};
+use openidconnect::{AdditionalProviderMetadata, ProviderMetadata};
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+struct RevocationEndpointProviderMetadata {
+    revocation_endpoint: String,
+}
+impl AdditionalProviderMetadata for RevocationEndpointProviderMetadata {}
+type GoogleProviderMetadata = ProviderMetadata<
+    RevocationEndpointProviderMetadata,
+    CoreAuthDisplay,
+    CoreClientAuthMethod,
+    CoreClaimName,
+    CoreClaimType,
+    CoreGrantType,
+    CoreJweContentEncryptionAlgorithm,
+    CoreJweKeyManagementAlgorithm,
+    CoreJwsSigningAlgorithm,
+    CoreJsonWebKeyType,
+    CoreJsonWebKeyUse,
+    CoreJsonWebKey,
+    CoreResponseMode,
+    CoreResponseType,
+    CoreSubjectIdentifierType,
+>;
+
 fn main() {
     env_logger::init();
 
@@ -55,11 +88,22 @@ fn main() {
         IssuerUrl::new("https://accounts.google.com".to_string()).expect("Invalid issuer URL");
 
     // Fetch Google's OpenID Connect discovery document.
-    let provider_metadata = CoreProviderMetadata::discover(&issuer_url, http_client)
+    // Note: if we don't care about token revocation we can simply use CoreProviderMetadata here
+    // instead of GoogleProviderMetadata.
+    let provider_metadata = GoogleProviderMetadata::discover(&issuer_url, http_client)
         .unwrap_or_else(|err| {
             handle_error(&err, "Failed to discover OpenID Provider");
             unreachable!();
         });
+
+    let revocation_endpoint = provider_metadata
+        .additional_metadata()
+        .revocation_endpoint
+        .clone();
+    println!(
+        "Discovered Google revocation endpoint: {}",
+        revocation_endpoint
+    );
 
     // Set up the config for the Google OAuth2 process.
     let client = CoreClient::from_provider_metadata(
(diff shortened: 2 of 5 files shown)


ximon18:
As the build fails unless `cargo fmt` is run (and produces that long line wrapping I mentioned) I'll just push `cargo fmt` anyway...

ximon18:
The test coverage test failed but I'm not sure what I can do about that. The tests seem to test only things specific o the openidconnect crate, not passthroughs to the oauth2 crate, which makes sense as those are tested in the oauth2 crate. Actual end-to-end functional testing seems only to be done by the ignored tests which I think were previously used with the openid.net self-certification platform. I don't think those conformance tests even cover aspects of OAuth 2 such as token introspection or revocation, even if I could amend the tests and run them.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "openidconnect"
-version = "2.0.0-alpha.3"
+version = "2.0.0-alpha.4"
```
ramosbugs, line 3: 2.0.0-alpha.3 still hasn't been released, so that's the correct version here I think

Cargo.toml
```diff
@@ -27,7 +27,7 @@ thiserror = "1.0"
 http = "0.2"
 itertools = "0.9"
 log = "0.4"
-oauth2 = { version = "4.0.0-alpha.3", default-features = false }
+oauth2 = { version = "4.0.0-alpha.4", default-features = false }
```
ramosbugs, line 30: mind pinning this to `=4.0.0-alpha.4`? it looks like the semver logic considers `4.0.0-alpha.4` to satisfy `4.0.0-alpha.3`, which is causing the current CI failures on the main branch. adding the `=` should prevent that going forward.

src/lib.rs
```diff
@@ -799,7 +821,7 @@ where
         G: GrantType,
         JK: JweKeyManagementAlgorithm,
         RM: ResponseMode,
-        RT: ResponseType,
+        RT2: ResponseType,
```
ramosbugs, line 824: nit: how about renaming to `RS` instead of `RT2`

src/lib.rs
```diff
@@ -833,6 +855,25 @@ where
...
+        self.oauth2_client = self.oauth2_client.set_introspection_url(introspection_url);
+        self
+    }
+
+    ///
+    /// Sets the revocation URL for contacting the revocation endpoint ([RFC 7009](https://tools.ietf.org/html/rfc7009)).
+    ///
+    /// See: [`revoke_token()`](Self::revoke_token())
+    ///
+    pub fn set_revocation_uri(mut self, revocation_uri: RevocationUrl) -> Self {
```
ramosbugs, line 872: thoughts on `uri` vs `url` here?

ramosbugs (commented):
thanks!

> cargo fmt has NOT been applied to the code as doing so causes

I'm not terribly bothered by the line breaks 🤷‍♂️

> I'm also unsure about error handling. In oauth2-rs if the revocation URL has not been set or is not HTTPS this was implemented such that it causes a TokenRequestError. However, I see that fn user_info() handles a similar case in a different way by returning a NoUserInfoEndpoint error.

ah, I had a thought about the delayed error when reviewing the oauth2 change but forgot about the `NoUserInfoEndpoint` trick I used in this crate. I think `NoUserInfoEndpoint` is probably a better pattern since the user gets the error sooner, but either approach seems ok.

> I also realize that I haven't run or updatd the tests or attempted to build with anything except default features.

don't worry about the test coverage here. I'm comfortable with the relevant tests living in the oauth2 crate, since this crate is just providing some plumbing.

Cargo.toml
ximon18: Done.

src/lib.rs
ximon18: Done.

Cargo.toml
ximon18: Done.

src/lib.rs
ximon18: I copied the existing pattern set by `fn set_redirect_uri()` although now I see that `fn set_revocation_url()` is inconsistent. RFC 6749 refers to redirection URI while RFC 7009 refers to revocation URL. RFC 7662 ducks the issue and refers only to the token endpoint defined by RFC 6749 which gives us 2/3 for URI and 1/3 for URL. URI can be either a URL or a URN, but in our case we are only talking about URLs, not URNs, because clients have to make HTTP requests to these endpoints and they can only make those requests to the URL subset of URIs, or even a subset of URLs (e.g. HTTPS and not HTTP). According to https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata the authorization_endpoint, token_endpoint and userinfo_endpoint are all URLs, not URIs.

I'd be inclined to standardize on URL.

ximon18: Regarding this:

> ah, I had a thought about the delayed error when reviewing the oauth2 change but forgot about the NoUserInfoEndpoint trick I used in this crate. I think NoUserInfoEndpoint is probably a better pattern since the user gets the error sooner, but either approach seems ok.

I think the NoUserInfoEndpoint approach is preferable as it not only gives earlier feedback to the developer but also the TokenRequestError based approach "hides" an input error in among many other possible types of error making it harder to take appropriate action in the calling code.

Should I therefore push a change to the oauth2-rs crate to adopt this style of error handling for the revocation and introspection URLs, adding errors for missing and invalid endpoint, and going further do we even need separate errors per endpoint (e.g. NoUserInfoEndpoint, NoIntrospectionEndpoint, NoRevocationEndpoint) or is it enough to just re-use an error such as MissingEndpointURL and InvalidEndpointURL ?

ramosbugs: > Should I therefore push a change to the oauth2-rs crate to adopt this style of error handling for the revocation and introspection URLs, adding errors for missing and invalid endpoint

if you'd like to, please go ahead! I'll happily merge this improvement.

> and going further do we even need separate errors per endpoint (e.g. NoUserInfoEndpoint, NoIntrospectionEndpoint, NoRevocationEndpoint) or is it enough to just re-use an error such as MissingEndpointURL and InvalidEndpointURL ?

using a common/shared error type seems fine. I'd suggest including a `&'static str` within it so that the `Display` output says which specific endpoint URL is missing, but I don't see any reason we need to differentiate between those errors in terms of static typing

ramosbugs: > I'd be inclined to standardize on URL.

sgtm

ximon18:
Working on the remaining items, hopefully will be able to push a PR tomorrow.

ximon18:
> Working on the remaining items, hopefully will be able to push a PR tomorrow

The changes to use a `NoUserInfoEndpoint` style error return in the `oauth2-rs` crate are more problematic than I expected.

1. It's not just the new functions for introspection and revocation that can fail due to a missing endpoint URL, existing functions can fail for this reason too and also currently return a _"late"_ `TokenRequestError::Other` error type, e.g. in `CodeTokenRequest`, `RefreshTokenRequest`, `PasswordTokenRequest`, `ClientCredentialsTokenRequest` and `DeviceAuthorizationRequest`.
2. Changing these functions causes a `Result` response in the call prior to `.request()` which wasn't there before which will require existing code to be changed to handle the `Result`.
3. When testing for the required URL earlier it makes no sense to then pass that URL as an `Option` to the later stage. However, while looking at removing this I realized that the existing code is not consistent, in some cases it passes a `&'a` reference, in others it passes a `Cow`. E.g. `Client::authorize_url()` does `redirect_url: self.redirect_url.as_ref().map(Cow::Borrowed)` while `Client::exchange_code()` does `redirect_url: self.redirect_url.as_ref()`.
4. The manner of providing configuration is not consistent, e.g. `fn DeviceAccessTokenRequest::request()` takes a `timeout` parameter which one can wonder whether that is really something that should be set per request or just once in the client, and this can cause `RequestTokenError::Other("Failed to convert duration".to_string())` or `RequestTokenError::Other("Failed to calculate timeout".to_string()))`.

The major inconsistency at present is the _"early"_ `NoUserInfoEndpoint` result, to be more consistent that could be changed to `RequestTokenError::Other` and become a _"late"_ result, rather than changing the new `introspect()` and `revoke_token()` functions in the `oauth2-rs` crate, but that would require current code to be changed to no longer check for a `Result` return value when calling `user_info()`.

(another minor inconsistency I just realized is that perhaps `fn introspect()` should really be called `fn introspect_token()`?)

@ramosbugs: How would you advise to proceed?

ramosbugs:
thank you for your amazing attention to detail!

> The changes to use a `NoUserInfoEndpoint` style error return in the `oauth2-rs` crate are more problematic than I expected.
> 
> 1. It's not just the new functions for introspection and revocation that can fail due to a missing endpoint URL, existing functions can fail for this reason too and also currently return a _"late"_ `TokenRequestError::Other` error type, e.g. in `CodeTokenRequest`, `RefreshTokenRequest`, `PasswordTokenRequest`, `ClientCredentialsTokenRequest` and `DeviceAuthorizationRequest`.
> 2. Changing these functions causes a `Result` response in the call prior to `.request()` which wasn't there before which will require existing code to be changed to handle the `Result`.

I remember thinking about this particular error handling pattern when I was originally designing this API, and I felt that while the token endpoint is technically optional and an OAuth2 provider can exist with only an authorization endpoint (using only the Implicit Flow), this is extremely rare in practice (I've never seen it). Since essentially every OAuth2 provider has a token endpoint, I didn't want to make these functions return a `Result` and force all callers to handle the extremely unlikely scenario. Instead, I just deferred that error case to `.request()`.

However, the same logic doesn't apply to other endpoints such as user info, token introspection, and token revocation. It's common for OAuth2 providers not to implement all of those, so it feels more reasonable to return a `Result` earlier, even if it means there's some inconsistency between the APIs that use the token endpoint and those that use other endpoints.

Does that make sense?

> 3. When testing for the required URL earlier it makes no sense to then pass that URL as an `Option` to the later stage. However, while looking at removing this I realized that the existing code is not consistent, in some cases it passes a `&'a` reference, in others it passes a `Cow`. E.g. `Client::authorize_url()` does `redirect_url: self.redirect_url.as_ref().map(Cow::Borrowed)` while `Client::exchange_code()` does `redirect_url: self.redirect_url.as_ref()`.

The inconsistency here seems less problematic since it's an implementation detail and not part of the public API. As long as we're avoiding unnecessary clones, feel free to use your best judgement here.

> 4. The manner of providing configuration is not consistent, e.g. `fn DeviceAccessTokenRequest::request()` takes a `timeout` parameter which one can wonder whether that is really something that should be set per request or just once in the client, and this can cause `RequestTokenError::Other("Failed to convert duration".to_string())` or `RequestTokenError::Other("Failed to calculate timeout".to_string()))`.

The [Device Authorization Flow](https://tools.ietf.org/html/rfc8628) involves repeatedly polling the server while waiting for the end user to take out-of-band action, so the timeout here is particularly important to the UX, vs. a less-often-customized HTTP timeout or similar config. I think that's why I [suggested](https://github.com/ramosbugs/oauth2-rs/pull/113#discussion_r479846150) passing it at request time. I've been trying to strike a balance between ergonomics and an API that clearly expresses the semantics of each operation, which explains some of the inconsistencies.

> The major inconsistency at present is the _"early"_ `NoUserInfoEndpoint` result, to be more consistent that could be changed to `RequestTokenError::Other` and become a _"late"_ result, rather than changing the new `introspect()` and `revoke_token()` functions in the `oauth2-rs` crate, but that would require current code to be changed to no longer check for a `Result` return value when calling `user_info()`. Although changing `NoUserInfoEndpoint` to something more generic to be re-used in multiple places would also impact code calling `user_info()`.

How would you feel about returning the same, e.g., `MissingEndpoint` error from `user_info()` and the other non-token-endpoint APIs, but leaving the token endpoint APIs unchanged?

> (another minor inconsistency I just realized is that perhaps `fn introspect()` should really be called `fn introspect_token()`?)

Either is fine with me.

> FYI: in my work in progress I implemented an error like so to be returned early. If you advise to continue down the path of using more early errors of a new generic type it might be good to get your input on this "design" already:
> 
> ```rust
> #[derive(Debug, thiserror::Error)]
> pub enum ConfigurationError {
>     #[error("No {0} endpoint URL specified")]
>     MissingURL(&'static str),
>     #[error("Scheme for {0} endpoint URL must be HTTPS")]
>     InsecureURL(&'static str),
> }
> ```

Looks good! I'd suggest making this `#[non_exhaustive]` so we can add to it in the future, and I'd also rename `URL` to `Url` in the enum variant names for consistency with the rest of the crate.

ximon18:
All sounds good, will try and finish the PR as soon as I can but might take a couple of days due to limited time.

ximon18:
Pushed. This branch will fail to compile until https://github.com/ramosbugs/oauth2-rs/pull/127 is merged.

ramosbugs:
Released https://crates.io/crates/oauth2/4.0.0-alpha.5 with your latest changes

ximon18:
Coming back to this I think it would be good to refine the implementation a little bit:

> RFC 6749 refers to redirection URI while RFC 7009 refers to revocation URL. RFC 7662 ducks the issue and refers only to the token endpoint defined by RFC 6749 which gives us 2/3 for URI and 1/3 for URL. URI can be either a URL or a URN, but in our case we are only talking about URLs, not URNs, because clients have to make HTTP requests to these endpoints and they can only make those requests to the URL subset of URIs, or even a subset of URLs (e.g. HTTPS and not HTTP). According to https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata the authorization_endpoint, token_endpoint and userinfo_endpoint are all URLs, not URIs.
> 
> I'd be inclined to standardize on URL.

I just found that `registration.rs` (which I guess implements [OpenID Connect Dynamic Client Registration 1.0](https://openid.net/specs/openid-connect-registration-1_0.html), not sure whether it incorporates errata set 1 or not) also refers to URIs, not URLs.

The [OpenID Connect Core 1.0 specification](https://openid.net/specs/openid-connect-core-1_0.html) adds more confusion to this as it defines "_urI" terms and documents them as "URLs"...

So the state of specifications and their references to URIs and/or URLs looks like this:

| Specification | Term Naming | Documentation Refers To |
|---|---|---|
| RFC 6749 | URI | URI |
| RFC 7009 | None | Delegates to RFC 6749 but also uses both URI and URL |
| RFC 7662 | None | Delegates to RFC 6749 |
| OpenID Connect Core 1.0 | URI | Delegates to RFC 6749 but also uses both URI and URL |
| OpenID Connect Discovery 1.0 | URI | Delegates to RFC 6749 but also uses both URI and URL |
| OpenID Connect Dynamic Client Registration 1.0 | URI | Delegates to RFC 6749 but also uses both URI and URL |

So, we have:
- Consistency in terms being named as URIs.
- Inconsistency in definitions of terms with both URIs and URLs being referenced.
- At an implementation level the OIDC and OAuth2 crates make HTTP(S) requests, they cannot connect to the [myriad other URI schemes](https://en.wikipedia.org/wiki/List_of_URI_schemes) that exist, e.g. how would it be sensible for the OpenID Connect crate to try to connect to an OAuth 2 token endpoint via an URI with scheme "ipps" (Internet Printing Protocol), or to a URI that uses a "bitcoin" URI scheme, or to a URI using the "chrome-extension" scheme?

I think it's most helpful to developers using these crates to clearly indicate that they are expected to work with URLs, not with URIs, but for ease of relating the crate APIs to the specs they implement perhaps terms should be named "_uri" consistent with the specs, as an example:

```rust
pub fn set_redirect_url(mut self, redirect_uri: RedirectUrl) -> Self {
        self.oauth2_client = self.oauth2_client.set_redirect_uri(redirect_uri);
        self
}
```

Here the type is specifically a URL, but the referenced spec field is named URI [per the spec](https://tools.ietf.org/html/rfc6749#appendix-A.6).

I think my earlier suggestion to standardize on URL was practical but perhaps should be refined to say that that we use the spec term/field names as they are defined (e.g. `_uri`) but use URL in crate type names as that is what the crates can actually use.

@ramosbugs: Thoughts?

ramosbugs:
> I think my earlier suggestion to standardize on URL was practical but perhaps should be refined to say that that we use the spec term/field names as they are defined (e.g. `_uri`) but use URL in crate type names as that is what the crates can actually use.
> 
> @ramosbugs: Thoughts?

sounds good to me! thanks for the thorough research and thoughtful analysis :-)

src/lib.rs
```diff
@@ -1025,6 +1048,36 @@ where
...
+    }
+
+    ///
+    /// Creates a request builder for revoking a previously received token.
+    ///
+    /// Requires that [`set_revocation_url()`](Self::set_revocation_url()) have already been called to set the
+    /// revocation endpoint URL.
+    ///
+    /// Attempting to submit the generated request without calling [`set_revocation_url()`](Self::set_revocation_url())
+    /// first will result in an error.
```
ramosbugs, line 1071: mind copying this comment (replaced with `set_introspection_url`) to `introspect()` above?

examples/google.rs
```diff
@@ -41,6 +39,41 @@ fn handle_error<T: std::error::Error>(fail: &T, msg: &'static str) {
     exit(1);
 }
 
+// Teach openidconnect-rs about the Google custom extension to the OpenID Discovery response
+// that we can use as the RFC 7009 OAuth 2.0 Token Revocation endpoint.
+use openidconnect::core::{
```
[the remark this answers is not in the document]
ximon18: I'm fine with moving these to the top but FYI I put them here to group everything together that was needed to handle non-standard extensions to the discovery response.

ximon18:
@ramosbugs : I've updated the PR as discussed, let me kow if you still see anything that needs tweaking or changing. I fixed a couple of other minor things as well such as removing an outdated "introspection TODO" comment and correcting my doc comment code to use hhe publically available API rather than the internal API to destructure the error response, and applied cargo fmt. This branch again depends on as yet unreleased changes pushed to the oauth2-rs crate repo, see https://github.com/ramosbugs/oauth2-rs/pull/128.

ramosbugs:
Released https://crates.io/crates/oauth2/4.0.0-alpha.6 with your latest changes

ximon18:
I thought I’d tested with all features but I see a failure in Travis. I’ll deal with that today.

ximon18:
@ramosbugs: tests pass now, apart from code coverage.

ramosbugs:
thanks!

PR #45, opened 2021-06-15
fix: disable openid scope

oficial:
Hi,
The 'disable_openid_scope' function are actually enabling the property.


ramosbugs:
nice catch, thank you!

paulocmoreno:
Hi @ramosbugs, do you have an estimate of when you will release this fix? Thanks.

ramosbugs:
I just cut a new release with this fix! https://crates.io/crates/openidconnect/2.1.0

PR #53, opened 2021-11-25
Parse `n` and `e` as BigInts

ubamrein:
Currently, signature verification fails, if a JWK contains a RSA public-key with a modulus, which has a leading zero. Though, according to [RFC7518](https://datatracker.ietf.org/doc/html/rfc7518#section-6.3.1.1) the serialized big integer should be unsigned (hence no need for the leading zero), it is also states:

>  Note that implementers have found that some cryptographic libraries
   prefix an extra zero-valued octet to the modulus representations they
   return, for instance, returning 257 octets for a 2048-bit key, rather
   than 256.  Implementations using such libraries will need to take
   care to omit the extra octet from the base64url-encoded
   representation.

This PR adds the `num_bigint` crate to take care of the big integers, to parse them as positive integers, and afterwards serialize them as unsigned big integers without leading zeros (as `ring` is [expecting](https://briansmith.org/rustdoc/ring/signature/struct.RsaPublicKeyComponents.html#structfield.n)).

Technically, in favour of not adding `num_bigint` as a dependency, one could probably just check for leading zeros and drop them accordingly. Personally I'd prefer the addition of the crate rather than dealing with big integers by myself.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ serde_path_to_error = "0.1"
 serde-value = "0.6"
 untrusted = "0.7"
 url = { version = "2.1", features = ["serde"] }
+num-bigint = "0.4.3"
 
 [dev-dependencies]
 color-backtrace = { version = "0.4" }
--- a/src/core/crypto.rs
+++ b/src/core/crypto.rs
@@ -1,3 +1,4 @@
+use num_bigint::{BigInt, Sign};
 use ring::hmac;
 use ring::rand::SecureRandom;
 use ring::signature as ring_signature;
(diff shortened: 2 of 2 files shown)


ramosbugs:
Thanks for the PR and the helpful reference citations!

> Currently, signature verification fails, if a JWK contains a RSA public-key with a modulus, which has a leading zero. Though, according to [RFC7518](https://datatracker.ietf.org/doc/html/rfc7518#section-6.3.1.1) the serialized big integer should be unsigned (hence no need for the leading zero), it is also states:
> 
> > Note that implementers have found that some cryptographic libraries
> > prefix an extra zero-valued octet to the modulus representations they
> > return, for instance, returning 257 octets for a 2048-bit key, rather
> > than 256.  Implementations using such libraries will need to take
> > care to omit the extra octet from the base64url-encoded
> > representation.

This section implies that it's the JWK serializer's responsibility to omit leading zeroes, and that JWKs that include them are malformed. However, it doesn't explicitly use language stating that the JWK "MUST NOT" include leading zeroes, so I think it's reasonable to have this crate tolerate them.

To avoid future regressions in this area, would you mind adding a short test case that fails verification before the change and passes after? This function includes existing RSA signing tests as examples: https://github.com/ramosbugs/openidconnect-rs/blob/f795bd3bc810663b4b18ec175d4676a737c4d40d/src/core/jwk.rs#L1123


Out of curiosity, have you seen any major/public OIDC providers that include leading zeroes?

ubamrein:
> To avoid future regressions in this area, would you mind adding a short test case that fails verification before the change and passes after? This function includes existing RSA signing tests as examples:

I did put the tests directly in the `crypto` module since it is not related to JWTs, but to the verification operation. This way I have trivial access to all private functions in this module, to actually test pre and post fix.

> Out of curiosity, have you seen any major/public OIDC providers that include leading zeroes?

We came across this problem while trying to use (an old?) version of [WSO2](https://wso2.com/identity-server/) identity provider. We actually don't control the IDP setup, so we don't really know, how the JWK was generated and serialised.

But we saw that e.g. the Java implementation had no problems verifying the JWT (as Java also internally uses BigIntegers), as well as the PHP version of `openssl_verify` (though to be fair, the JWK was converted to a `SubjectPublicKeyInfo` PEM file).

Interestingly enough [jwt.io](https://jwt.io) also cannot handle public keys in JWK format with leading zeros, which did not help while debugging the issue :D

ramosbugs:
thank you! this is now released in version 2.1.2

PR #55, opened 2021-12-22
Add optional feature to accept RFC 3339 timestamps, for compatibility with Auth0

### LINKED ISSUE #23: Unable to parse `updated_at` timestamp from Auth0
I'm working on integrating this library with Auth0 and ran into an issue where the format of their `updated_at` field fails to parse. Their format is `"2020-06-16T03:23:10.426Z"`.

I changed this type to `String` in the `claims.rs` and it parses successfully.

andrewhickman:
Fixes #23

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ curl = ["oauth2/curl"]
 reqwest = ["oauth2/reqwest"]
 native-tls = ["oauth2/native-tls"]
 rustls-tls = ["oauth2/rustls-tls"]
+accept-rfc3339-timestamps = []
 nightly = []
 
 [dependencies]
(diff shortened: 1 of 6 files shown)


ramosbugs:
Thank you!

ThomasLamprecht:
Is there any reason that this is an (optional) feature? It doesn't pull in any new dependency FWICT and runtime overhead of that additional match shouldn't be too bad, and users of this crate would benefit from working with more providers by default.

I mean, I'll enable it and be done for my use case, so no biggie for me personally, but after reading the release notes and upgrading did not solve auth0 access I had to read more closely and check the PR itself, I wondered for the reason behind that feature guard.

ramosbugs:
it's optional because Auth0 does not follow the OpenID Connect standard, and this library adheres to the standard in its default configuration

PR #56, opened 2021-12-22
Expose ureq feature of oauth2

andrewhickman:


andrewhickman:
```
error[E0658]: use of unstable library feature 'bool_to_option'
   --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/rustls-0.20.2/src/tls13/mod.rs:109:58
    |
109 |         (prev.hash_algorithm() == self.hash_algorithm()).then(|| prev)
    |                                                          ^^^^
```

I think the CI failure is because the MSRV of the latest rustls ([1.52.1](https://github.com/rustls/rustls/pull/728)) is higher than we're using in CI.

It looks like reqwest is picking up the same rustls version so I'd expect the main branch CI to fail as well

ramosbugs:
> I think the CI failure is because the MSRV of the latest rustls ([1.52.1](https://github.com/rustls/rustls/pull/728)) is higher than we're using in CI.
> 
> It looks like reqwest is picking up the same rustls version so I'd expect the main branch CI to fail as well

I think the issue is a stale Cargo-1.45.lock that needs to be updated to include a version of ureq that's compatible with 1.45. The CI for `oauth2-rs` is passing on 1.45 when pinned to ureq 2.0.1, so let's see if updating the lockfile will make CI pass.

ramosbugs:
success, thanks for the PR!

PR #59, opened 2022-01-08
Implement `FromStr` for `IdToken`

### LINKED ISSUE #57: Implement `FromStr` for `IdToken`
We want to accept an [`IdToken`](https://docs.rs/openidconnect/2.1.2/openidconnect/struct.IdToken.html) from our clients (e.g. a cli that interacts with our REST API). For that, they should send their id token via a header (e.g. `Authorization` header with `Bearer` type). However, I've searched in the documentation and source code and can't find a way to get an `IdToken` from a String. The IdToken would be provided in the usual JWT Format (base64 encoded `header.claims.signature`). Am I missing something? If so, could you provide an example?

Erik1000:
Fixes #57

### DIFF
--- a/src/id_token.rs
+++ b/src/id_token.rs
@@ -434,6 +434,7 @@ where
 #[cfg(test)]
 mod tests {
     use std::collections::HashMap;
+    use std::str::FromStr;
 
     use chrono::{TimeZone, Utc};
     use oauth2::basic::BasicTokenType;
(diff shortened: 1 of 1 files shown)


ramosbugs:
thanks!

ramosbugs:
This is now released in https://crates.io/crates/openidconnect/2.2.0

PR #61, opened 2022-01-22
Add insecure_unverified_claims method

gautamg795:
Adds a method to read out the claims of a token without key verification.
Our use case is that there may be multiple whitelisted identity providers configured for our application, and rather than fetching the metadata for each one then attempting verification with each JWK set, we'd like to read out the issuer from the token and see if any of our whitelisted IDPs match. 

I think it's reasonable to provide an insecure escape hatch for situations like this, but I'd be open to any alternative solutions you might have!

### DIFF
--- a/src/id_token.rs
+++ b/src/id_token.rs
@@ -126,11 +126,6 @@ where
         verifier.verified_claims(&self.0, nonce_verifier)
     }
 
-    /// Returns a reference to the token's claims **without performing any verification of the token's signature.**
-    pub fn insecure_unverified_claims<'a>(&'a self) -> &'a IdTokenClaims<AC, GC> {
-        self.0.unverified_payload_ref()
-    }
-
     ///
     /// Verifies and returns the ID token claims.
     ///
--- a/src/id_token.rs
+++ b/src/id_token.rs
@@ -126,6 +126,11 @@ where
         verifier.verified_claims(&self.0, nonce_verifier)
     }
 
+    /// Returns a reference to the token's claims **without performing any verification of the token's signature.**
+    pub fn insecure_unverified_claims<'a>(&'a self) -> &'a IdTokenClaims<AC, GC> {
+        self.0.unverified_payload_ref()
+    }
+
     ///
     /// Verifies and returns the ID token claims.
     ///
--- a/src/verification.rs
+++ b/src/verification.rs
@@ -553,6 +553,23 @@ where
     }
 
     ///
+    /// Initializes a no-op verifier that performs no signature, audience, or issuer verification.
+    /// The token's expiration time is still checked, and the token is otherwise required to conform to the expected format.
+    ///
+    pub fn new_insecure_without_verification() -> Self {
+        let empty_issuer = IssuerUrl::new("https://0.0.0.0".to_owned())
+            .expect("Creating empty issuer url mustn't fail");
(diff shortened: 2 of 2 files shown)


ramosbugs:
Hey @gautamg795,

The use case you described makes sense to me, but I'd like to make sure that the existing APIs don't already satisfy it before adding a new escape hatch. Can you call the normal `IdToken::claims` method but provide it with an `IdTokenVerifier` that ignores the signature? There's an [IdTokenVerifier::insecure_disable_signature_check](https://docs.rs/openidconnect/2.2.0/openidconnect/struct.IdTokenVerifier.html#method.insecure_disable_signature_check) method that I think should help. Does that work for you?

ramosbugs:
You'll probably need to disable the issuer check as well: https://docs.rs/openidconnect/2.2.0/openidconnect/struct.IdTokenVerifier.html#method.require_issuer_match

gautamg795:
yeah, I think there is a way to make that work — would need to disable the audience/issuer/signature checks, as well as pass a no-op nonce verifier. it felt unergonomic enough that it was worth having a more convenient method, but I could certainly understand an argument that this _should_ be difficult

ramosbugs:
> it felt unergonomic enough that it was worth having a more convenient method, but I could certainly understand an argument that this should be difficult

exactly! given the number of StackOverflow posts that suggest unwise things like disabling TLS certificate verification as a solution to signature errors, I'd prefer that the user have to make it very explicit in their code that they know what they're doing and that they want to do it anyway 😀

gautamg795:
the only thing I particularly don't like about that solution is that the `CoreIdTokenVerifier::new_public_client` call needs to be made with a dummy issuer, dummy client id, and empty JWKS. having to make a dummy verifier with fake values in production code feels pretty icky to me. 

would this escape hatch be more palatable if it were hidden behind a clearly-labeled trait that needs to be imported for the method to be called?

ramosbugs:
what if we add a new factory method for `IdTokenVerifier` called `new_insecure_without_verification` that avoids having to pass in any dummy values? that would still leave the top-level `IdToken` interface unchanged.

gautamg795:
yeah, I think that works for me. will update this PR soon

gautamg795:
Updated to match your suggestion, let me know what you think.

ramosbugs:
looks good! I think ideally the tests should also cover each of the cases where this verifier disables a check. if someone were to comment out any of the following lines, a test should fail since it'll potentially break functionality that someone will rely on:
https://github.com/ramosbugs/openidconnect-rs/blob/8123e0117209c8d497fb765a136dcebe7ddfe1ba/src/verification.rs#L567-L569

I think the existing tests `test_jwt_verified_claims`/`test_id_token_verified_claims` cover these cases for the secure verifier, so it's just a matter of adding a call using the new verifier and making sure it doesn't fail.

gautamg795:
just rebased on your main and added the calls to the insecure verifier as requested!

gautamg795:
thanks for the guidance on this one! let me know when you've had a chance to release (no rush) so I can move my dependent crates back to the main release

gautamg795:
hey @ramosbugs, could I request a release to crates.io including this change?

ramosbugs:
sure, just released https://crates.io/crates/openidconnect/2.3.0

PR #66, opened 2022-03-03
Fix `ClientRegistrationResponse` failing to deserialize optional fiel…

sunli829:
This PR fixes the following two issues:

## Fix `ClientRegistrationResponse` failing to deserialize optional fields `client_id_issued_at` and `client_secret_expires_at`.

> By default, serde accepts a missing field if the type of the field is Option<_>. If you add a with or deserialize_with attribute serde will fail with the "missing field" error message, as you noticed. This is on purpose. You can change what happens on missing fields with the default attribute, either applied to the field in question or the whole struct. 

Reference: https://github.com/serde-rs/serde/issues/1951

## Made `XXXUrl`'s equal behavior consistent with `url::Url`.

The following two Urls are now equal.

```rust
assert_eq!(
    IssueUrl::new("http://example.com").unwrap(),
    IssueUrl::new("http://example.com/").unwrap(),
);
```

https://github.com/ramosbugs/openidconnect-rs/blob/d5cf0ae0c80893f060157225b77284710d481291/src/discovery.rs#L404

### DIFF
--- a/src/registration.rs
+++ b/src/registration.rs
@@ -698,12 +698,14 @@ where
     registration_client_uri: Option<ClientConfigUrl>,
     #[serde(
         skip_serializing_if = "Option::is_none",
-        with = "serde_utc_seconds_opt"
+        with = "serde_utc_seconds_opt",
+        default
     )]
     client_id_issued_at: Option<DateTime<Utc>>,
     #[serde(
         skip_serializing_if = "Option::is_none",
-        with = "serde_utc_seconds_opt"
+        with = "serde_utc_seconds_opt",
+        default
     )]
     client_secret_expires_at: Option<DateTime<Utc>>,
     #[serde(bound = "AC: AdditionalClientMetadata", flatten)]


ramosbugs:
Thanks for the PR!

The URL exact string comparison is intended behavior, as documented here:
https://github.com/ramosbugs/openidconnect-rs/blob/d5cf0ae0c80893f060157225b77284710d481291/src/macros.rs#L258-L261

From the [spec](https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation):
> 2. The Issuer Identifier for the OpenID Provider (which is typically obtained during Discovery) MUST exactly match the value of the `iss` (issuer) Claim.

I don't see any mention of URL canonicalization in the spec, and various forms of canonicalization have caused security vulnerabilities in other authentication protocols like SAML. Consequently, I think it makes sense to stick to exact URL string comparisons throughout this crate. Client code is free to do a looser comparison if desired by first calling the `.url()` method and then comparing the returned `&Url`.

I'm happy to merge the deserialization fix though. Nice catch 👍

sunli829:
Thanks for you quick response! 😁

Using `https://broker.pod.inrupt.com` to call the `CoreProviderMetadata::discover_async` function will trigger an error, the error output is:

````
unexpected issuer URI `https://broker.pod.inrupt.com/` (expected `https://broker.pod.inrupt.com/`)
````

Need to use `https://broker.pod.inrupt.com/` to make sure the url is the same as defined by `openid-configuration`, which is inconvenient, don't you think? 😁

ramosbugs:
> Using `https://broker.pod.inrupt.com` to call the `CoreProviderMetadata::discover_async` function will trigger an error, the error output is:
> 
> ```
> unexpected issuer URI `https://broker.pod.inrupt.com/` (expected `https://broker.pod.inrupt.com/`)
> ```

Ah, I see! I think we can improve the error output here to print the `.as_str()` output instead, so that at least the error message matches the semantics of the comparison:
https://github.com/ramosbugs/openidconnect-rs/blob/cfa5af581ee100791f68bf099dd15fa3eb492c8b/src/discovery.rs#L407-L408
(wonder if this issue exists elsewhere in the crate...)

> Need to use `https://broker.pod.inrupt.com/` to make sure the url is the same as defined by `openid-configuration`, which is inconvenient, don't you think? 😁

It doesn't involve any extra code, so I don't think it's inconvenient; just how OpenID Connect works. OIDC providers should consider this when picking their issuer URL.

sunli829:
Thanks a lot, I updated the PR. 😁

Issue #68, opened 2022-03-25
Support parsing logout tokens


Erik1000:
I want to implement [back-channel logout](https://openid.net/specs/openid-connect-backchannel-1_0.html). For this, the OP sends an so called Logout token to the RP. The RP needs to validate this token. Since it is not equal to an IdToken, it currently can't be parsed by this library.

Erik1000:
@ramosbugs would you mind linking #70 (or the corresponding branch) to this issue?

ramosbugs:
> @ramosbugs would you mind linking #70 (or the corresponding branch) to this issue?

Sure, I think just referencing it in the PR message accomplishes that

PR #69, opened 2022-03-25
Cleanup

Erik1000:
I wanted to contribute to this crate (#68) and while attempting to do so, rust-analyzer (with clippy) complaint. I thought, I could use this chance to fix the clippy errors and do some clean up.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,6 +25,7 @@ jobs:
     runs-on: ubuntu-latest
 
     strategy:
+      fail-fast: false
       matrix:
         rust:
           - 1.45.0
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,10 @@ nightly = []
 [dependencies]
 base64 = "0.13"
 # Disable 'time' dependency since it triggers RUSTSEC-2020-0071 and we don't need it.
-chrono = { version = "0.4", default-features = false, features = ["clock", "std"] }
+chrono = { version = "0.4", default-features = false, features = [
+    "clock",
+    "std",
+] }
 thiserror = "1.0"
 http = "0.2"
 itertools = "0.9"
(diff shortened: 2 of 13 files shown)


Erik1000:
CI fails because of ring (https://github.com/rust-lang/rust/issues/95267)

ramosbugs:
Thanks for the cleanup! I'll take a look once nightly gets fixed (sounds like early next week). My main concern about fixing clippy lints is that newer versions of clippy may suggest changes involving newer language features that break compatibility with the MSRV 1.45.0. As long as CI passes, the changes should be fine though.

Also, feel free to add [`fail-fast: false`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast) to the GitHub Actions job if you don't want to wait for the nightly fix. I think if I add that to the `main` branch right now it won't get pulled in unless you merge or rebase.

Erik1000:
```rust
error: use of `unwrap_or` followed by a function call
  --> src/http_utils.rs:17:31
   |
17 |             ct[..ct.find(';').unwrap_or(ct.len())].to_lowercase() == expected_essence.to_lowercase()
   |                               ^^^^^^^^^^^^^^^^^^^ help: try this: `unwrap_or_else(|| ct.len())`
   |
   = note: `-D clippy::or-fun-call` implied by `-D warnings`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#or_fun_call
   ```

We now have the situation that the clippy version from MSVC does not recognize that `ct.len()` is a const function (it is since 1.39, MSVC is 1.45) and therefore thinks that this is `unwrap_or` with a function call. The latest clippy however does recognize that and warns when using `unwrap_or_else` since it is not needed:
```rust
warning: unnecessary closure used to substitute value for `Option::None`
  --> src/http_utils.rs:17:18
   |
17 |             ct[..ct.find(';').unwrap_or_else(|| ct.len())].to_lowercase()
   |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use `unwrap_or` instead: `ct.find(';').unwrap_or(ct.len())`
   |
   = note: `#[warn(clippy::unnecessary_lazy_evaluations)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_lazy_evaluations
```

Using `unwrap_or` is -- no matter the compiler version -- the right choice, but what should we do about the false positive with the old clippy version?

Erik1000:
Related to https://github.com/rust-lang/rust-clippy/issues/6943

ramosbugs:
I think let's go with `unwrap_or` and annotate the line with `#[allow(clippy::or_fun_call)]` so that clippy 1.45.0 doesn't complain. that should make CI happy (other than the known ring issue on nightly).

ramosbugs:
thanks!

PR #70, opened 2022-03-27
Backchannel logout

Erik1000:


Erik1000:
This branch contains the commits from #69

ramosbugs:
now that I merged #69, mind rebasing and force pushing (or merging `main` back into this branch)? that should make it easier to review

src/lib.rs
```diff
@@ -596,6 +596,12 @@ pub use oauth2::{
     TokenResponse as OAuth2TokenResponse, TokenType, TokenUrl,
 };
 
+#[cfg(feature = "backchannel-logout")]
+mod logout_token;
+#[cfg(feature = "backchannel-logout")]
+#[doc(inline)]
+pub use logout_token::*;
```
Erik1000, line 603: This might actually be not such a good idea, I'll remove that. Otherwise, it might get confusing because of stuff like `Identifier`

src/backchannel_logout.rs
```diff
@@ -0,0 +1,469 @@
...
+    pub fn events(&self) -> &HashMap<String, serde_json::Value> {
+        &self.events
+    }
+}
+impl<'de> Deserialize<'de> for LogoutToken {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let value: serde_json::Value = serde_json::Value::deserialize(deserializer)?;
```
ramosbugs, line 76: If I'm reading the spec correctly, I think this struct should be parsed as a JWT rather than raw JSON, using similar code to `IdToken`:
> OPs send a JWT similar to an ID Token to RPs called a Logout Token to request that they log out.

The code should also verify the JWT signature:
> A Logout Token MUST be signed and MAY also be encrypted

I'd suggest structuring the code similarly to what this crate does for ID tokens:
- `LogoutTokenClaims` struct similar to `IdTokenClaims` (implementing both `Deserialize` and `Serialize` so that both the RP and OP code paths are supported)
- `LogoutToken` struct similar to `IdToken` that wraps `JsonWebToken`
- `LogoutTokenVerifier` struct similar to `IdTokenVerifier` or `UserInfoVerifier` that contains a `JwtClaimsVerifier` and makes sure that a `nonce` is omitted. I think I would just add an optional `nonce` field (annotated with `serde(skip_serializing)` so that it doesn't get serialized by an OP) to `LogoutTokenClaims` have the verifier check that it's `None`. There are some other required verification steps in the [spec](https://openid.net/specs/openid-connect-backchannel-1_0.html#Validation), including for `iat` (see `IdTokenVerifier::iat_verifier_fn`). `JwtClaimsVerifier` will take care of `iss` and `aud`.

src/backchannel_logout.rs
```diff
@@ -0,0 +1,469 @@
...
+        if let serde_json::Value::Object(ref map) = value {
+            if map.contains_key("nonce") {
+                return Err(<D::Error as Error>::custom("nonce claim is prohibited"));
+            }
+        }
+
+        #[derive(Deserialize)]
+        struct Repr {
```
ramosbugs, line 84: Following the suggestions above, I think the duplicate `Repr` and custom `Deserialize` implementation can probably be removed in favor of serde annotations where needed

src/backchannel_logout.rs
```diff
@@ -0,0 +1,469 @@
...
+            aud: Vec<Audience>,
+            /// Time at which this token was issued
+            #[serde(with = "serde_utc_seconds")]
+            iat: DateTime<Utc>,
+            /// The unique identifier for this token. This can be used to detect
+            /// replay attacks.
+            jti: String,
+            #[serde(flatten)]
+            identifier: Identifier,
+            events: HashMap<String, serde_json::Value>,
```
ramosbugs, line 98: Using types, we can have serde check the URL key for us:
```rust

#[derive(Clone, Debug, Deserialize, Serialize)]
#[serde(deny_unknown_fields)]
struct BackchannelLogoutEvent {}

#[derive(Clone, Debug, Deserialize, Serialize)]
struct LogoutTokenEvents {
  #[serde(rename = "http://schemas.openid.net/event/backchannel-logout")]
  backchannel_logout: BackchannelLogoutEvent,
}
```

src/backchannel_logout.rs
```diff
@@ -0,0 +1,469 @@
...
+            identifier: token.identifier,
+            events: token.events,
+        })
+    }
+}
+
+/// A [`LogoutToken`] MUST contain either a `sub` or a `sid` claim and MAY
+/// contain both. This enum represents these three possibilities.
+#[derive(Debug, Hash, Clone, PartialEq, Eq)]
+pub enum Identifier {
```
ramosbugs, line 128: I like this enum since it ensures that we can't represent a token with both fields set to `None`! I would just suggest naming it something a bit more descriptive like `LogoutIdentifier`.

src/backchannel_logout.rs
```diff
@@ -0,0 +1,469 @@
...
+    /// The audience this token is intended for
+    aud: Vec<Audience>,
+    /// Time at which this token was issued
+    iat: DateTime<Utc>,
+    /// The unique identifier for this token. This can be used to detect
+    /// replay attacks.
+    jti: String,
+    identifier: Identifier,
+    events: HashMap<String, serde_json::Value>,
+}
```
ramosbugs, line 33: Let's add a flattened generic `additional_claims` struct here too, similar to what the other token types do. This will allow users to define their own extensions. The spec says:
> Logout Tokens MAY contain other Claims. Any Claims used that are not understood MUST be ignored.

As long as the user's code understands the additional fields, then I think it's fine for them to be present.

I think this will need to use `FilteredFlatten` (see `IdTokenClaims::additional_claims` for an example) to filter out `sid` and `sub` since we're using multiple flattened fields and `Identifier`'s field names aren't known to serde's derive macro at compile time).

src/lib.rs
```diff
@@ -596,6 +596,9 @@ pub use oauth2::{
     TokenResponse as OAuth2TokenResponse, TokenType, TokenUrl,
 };
 
+#[cfg(feature = "backchannel-logout")]
+pub mod backchannel_logout;
```
ramosbugs, line 600: `pub mod` seems like the right approach since OIDC supports multiple logout mechanisms and, as you pointed out, it could get confusing to put everything in the top-level namespace.

Mind moving this down to where the other pub mods (e.g.,`core`) are declared?

src/backchannel_logout.rs
```diff
@@ -0,0 +1,469 @@
...
+pub struct LogoutToken {
+    /// The issuer of this token
+    iss: IssuerUrl,
+    /// The audience this token is intended for
+    aud: Vec<Audience>,
+    /// Time at which this token was issued
+    iat: DateTime<Utc>,
+    /// The unique identifier for this token. This can be used to detect
+    /// replay attacks.
+    jti: String,
```
ramosbugs, line 30: For consistency with the rest of the crate, let's define a `JsonWebTokenId` newtype instead of using a plain string. That'll also help if this crate ever needs to support JWT-based client auth.

ramosbugs (commented):
Thanks for the PR!

As part of this feature, I think it's worth adding the `backchannel_logout_supported` and `backchannel_logout_session_supported` fields to [`ProviderMetadata`](https://docs.rs/openidconnect/2.2.0/openidconnect/struct.ProviderMetadata.html) (per [Section 2.1](https://openid.net/specs/openid-connect-backchannel-1_0.html#BCSupport)). Since the contents of that struct are private and we only need to add new getters and setters, this change should be backward compatible.

Similarly, the spec defines `backchannel_logout_uri` and `backchannel_logout_session_required` for [`ClientMetadata`](https://docs.rs/openidconnect/2.2.0/openidconnect/registration/struct.ClientMetadata.html).

Cargo.toml
```diff
@@ -15,14 +15,15 @@ all-features = true
...
-default = ["reqwest", "rustls-tls"]
+default = ["reqwest", "rustls-tls", "backchannel-logout"]
 curl = ["oauth2/curl"]
 reqwest = ["oauth2/reqwest"]
 ureq = ["oauth2/ureq"]
 native-tls = ["oauth2/native-tls"]
 rustls-tls = ["oauth2/rustls-tls"]
 accept-rfc3339-timestamps = []
 nightly = []
+backchannel-logout = []
```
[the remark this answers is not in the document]
Erik1000: I just thought that since it is still a draft and breaking changes could happen, this way, there's no need for a major bump if we change something there (according to semver)

src/backchannel_logout.rs
Erik1000: > If I'm reading the spec correctly, I think this struct should be parsed as a JWT rather than raw JSON

You're totally right and it's because I planned on implementing this a bit differently than `IdToken` but guess I'll better keep the style of this crate then.

src/backchannel_logout.rs
Erik1000: Yes, I thought about this and came to the conclusion that it's not a good idea if not done properly. But in order to to this right, I guess we need to completely implement the [SET](https://www.rfc-editor.org/info/rfc8417) standard and have the different events typed up in something like an enum. Also, I think this

```rust
#[derive(Clone, Debug, Deserialize, Serialize)]
#[serde(deny_unknown_fields)] // <--- would violate the spec
struct BackchannelLogoutEvent {}
```

since the spec says about this ([Section 2.4](https://openid.net/specs/openid-connect-backchannel-1_0.html#LogoutToken)): 
> The corresponding member value MUST be a JSON object and **SHOULD** be the empty JSON object {}. 

it only SHOULD not MUST be empty, so we can't deny unknown fields, can we?

I've only added the check [here](https://github.com/ramosbugs/openidconnect-rs/pull/70/commits/091b675b0322154b8d5c0ddf0d3324589bc2c344#diff-94f740fa2f6aa14eea6873b596009047af7f9caef16d6ab80a619358233a7c7cR106) to ensure that we only parse tokens that are valid to the spec. I've even considered completely ignoring this field apart from making sure it complies to the spec.

Cargo.toml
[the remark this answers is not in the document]
ramosbugs: ah gotcha. I think I'd prefer to have a full working implementation in this PR before merging it so that it's ready to release at that point

src/backchannel_logout.rs
ramosbugs: Oops, you are correct! Let's remove `deny_unknown_fields` then. What if we leave the `BackchannelLogoutEvent` struct empty for now but use `#[non_exhaustive]` so that we can add fields in the future without a breaking change?

src/backchannel_logout.rs
ramosbugs: Hopefully it sounds like more of a pain that it'll actually be to implement. I've added various utility types (e.g., `FilteredFlatten`) for doing this in other places without much new code each time. This will be the third instance (after ID tokens and user info responses) of having extensible JWTs.

Cargo.toml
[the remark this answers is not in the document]
Erik1000: Well that would depend on the OpenID working group to finalise the specification

Erik1000:
Also, I think we should add support for the `sid` claim in the `IdToken` too (even tho this would be already possible with the additional claims support). The backchannel spec ([Section 2.1](https://openid.net/specs/openid-connect-backchannel-1_0.html#BCSupport)) links to the front channel spec for this: https://openid.net/specs/openid-connect-frontchannel-1_0.html#OPLogout
It's basically an additional claim in the `IdToken` which would use the already existing `SessionIdentifier` type.

src/backchannel_logout.rs
```diff
@@ -0,0 +1,495 @@
...
+where
+    AC: AdditionalClaims,
+{
+    fn should_include(field_name: &str) -> bool {
+        !matches!(
+            field_name,
+            "iss" | "aud" | "iat" | "jti" | "sub" | "sid" | "events"
+        )
+    }
+}
```
Erik1000, line 47: I don't think we need support for localization or something like that, right?

src/backchannel_logout.rs
Erik1000: Wouldn't it make sense to use a enum for `LogoutTokenEvents`?

src/backchannel_logout.rs
Erik1000: I'm not feeling so good implementing an extra `LogoutTokenVerifier` because it's basically copy-paste except for the `nonce` part. Maybe there's a way to be more generics about this?

src/backchannel_logout.rs
```diff
@@ -0,0 +1,543 @@
...
+    /// Time at which this token was issued
+    #[serde(with = "serde_utc_seconds")]
+    iat: DateTime<Utc>,
+    /// The unique identifier for this token. This can be used to detect
+    /// replay attacks.
+    jti: JsonWebKeyId,
+    #[serde(flatten)]
+    identifier: LogoutIdentifier,
+    events: HashMap<String, serde_json::Value>,
+    additional_claims: FilteredFlatten<Self, AC>,
```
ramosbugs, line 62: I think this should be:
```rust
additional_claims: FilteredFlatten<LogoutIdentifier, AC>,
```

The issue we're trying to solve here is closely related to serde-rs/serde#1346. If a struct contains any fields with `#[serde(flatten)]`, then Serde's derive macros treat the entire struct as a map rather than a "struct" in Serde's [data model](https://serde.rs/data-model.html#types).

The way deserialization works in this case is that all of the known field names (e.g., `iss`, `aud`, etc. in `LogoutTokenClaims`) are deserialized to the corresponding fields in the "outer" struct (`LogoutTokenClaims`), and all of the other key-value pairs from the input are gathered into a vec called `__collect`: https://github.com/serde-rs/serde/blob/fb2fe409c8f7ad6c95e3096e5e9ede865c8cfb49/serde_derive/src/de.rs#L2459-L2464.

Then, the key-value pairs in `__collect` are passed to the deserializers for each of the flattened fields. However, Serde doesn't do anything to ensure mutual exclusivity between flattened fields. This means that even though `LogoutIdentifier` will use `sub` and/or `sid`, those fields will _also_ be passed to `additional_claims`'s deserializer. The `FilteredFlatten` struct's job here is to filter out `sub` and `sid` so that they don't end up duplicated in `additional_claims`.

The [`cargo expand`](https://github.com/dtolnay/cargo-expand) output is really helpful for debugging serde behavior.

For more context, see https://github.com/ramosbugs/openidconnect-rs/commit/02c606445375b5b365ad466c96c19f2016aaaf26, which is where I originally fixed this issue for ID tokens and user info claims.

src/backchannel_logout.rs
ramosbugs: I don't see any mentions of that in the spec, so hopefully not :-)

src/backchannel_logout.rs
ramosbugs: hmm... I tried to move all of the shared JWT verification logic to `JwtClaimsVerifier`. The duplicated part here would mostly be the factory methods (e.g., `new_public_client`) and getters/setters I think. I'm not sure how to get around that extra boilerplate. There shouldn't be too much duplicated _logic_ though, other than verifying `iat`, which is simple enough.

I think I'd start by implementing this verifier with all of the needed functionality, and then maybe it'll be clear whether there are any opportunities to factor out common code.

src/backchannel_logout.rs
ramosbugs: Enums seem appropriate when we need to distinguish between mutually-exclusive types. In this case, don't we know that it should always contain an event with the `http://schemas.openid.net/event/backchannel-logout` key? What other info do you envision needing to represent with `LogoutTokenEvents`?

If extensibility is the issue, we could add a generic flattened `additional_claims` field. Since there aren't any other flattened fields here, we won't need `FilteredFlatten`.

Similarly, we could make `backchannel_logout` a generic type instead of defining it as the empty `BackchannelLogoutEvent` struct. It depends how extensible we want to get and what future functionality users might want without needing to introduce breaking changes.

ramosbugs:
> Also, I think we should add support for the `sid` claim in the `IdToken` too (even tho this would be already possible with the additional claims support). The backchannel spec ([Section 2.1](https://openid.net/specs/openid-connect-backchannel-1_0.html#BCSupport)) links to the front channel spec for this: https://openid.net/specs/openid-connect-frontchannel-1_0.html#OPLogout It's basically an additional claim in the `IdToken` which would use the already existing `SessionIdentifier` type.

That's worth considering. Feel free to file a separate issue for tracking that. I think it'll be a breaking change though since some users may be relying on `additional_claims` already to parse that field, and with this change it would no longer be passed down. It might be simpler to provider helper types that users can pass via `additional_claims` to avoid breaking backward compatibility for the `IdTokenClaims` struct itself.

Erik1000:
> That's worth considering. Feel free to file a separate issue for tracking that. I think it'll be a breaking change though since some users may be relying on additional_claims already to parse that field, and with this change it would no longer be passed down.

Yup, I thought exactly about that because I implemented that using an additional claim in one of my projects.

src/backchannel_logout.rs
Erik1000: The thing with this is that it might include other "events" than `http://schemas.openid.net/event/backchannel-logout` 
and as much as I'd like to have all of these typed in an enum or something, it's out of scope for this crate.

I think just exposing the HashMap like it is currently should be fine since I don't think anyone would actually care about this and even if someone does, they could just use `serde_json::from_value` to parse their structs.

> What other info do you envision needing to represent with LogoutTokenEvents?

Forget that, I thought we could do something like this, but that ain't gonna work:

```rust
enum Event {
    #[serde(rename = "http://schemas.openid.net/event/backchannel-logout")]
    Backchannel {},
    // other possible SET event
    #[serde(rename = "https://schemas.openid.net/secevent/risc/event-type/account-disabled")]
    AccountDisabled {
         reason: String
    }
}
```

src/backchannel_logout.rs
Erik1000: Alright then, I'll see what I can do tomorrow.

KevinNaidoo:
Hi @Erik1000 / @ramosbugs,

I'd like to move this ahead, I'm currently implementing backchannel logout and have started adding functionality to the work done here. Any concerns around the approach or has development stopped because of other work?

Erik1000:
> Hi @Erik1000 / @ramosbugs,
> 
> I'd like to move this ahead, I'm currently implementing backchannel logout and have started adding functionality to the work done here. Any concerns around the approach or has development stopped because of other work?

Hi, work here has stopped because the JWT implementation has some issues (including missing support for JWE and it requires some unnecessary code duplication). A friend of mine and me have started a JOSE implementation that tries to solve these problems with a better, more extensible, and powerful way including an rust only crypto backend (which solves #58). We’re aiming for our first stable release within the next two weeks (see <https://github.com/minkan-chat/jose/issues/30> and also look at the other branches). Current state is that JWS is working with a great api and we just want to implement an better way for payloads (needed because of detached content). 
When we’re ready, back channel tokens can easily be parsed.

KevinNaidoo:
> > Hi @Erik1000 / @ramosbugs,
> > I'd like to move this ahead, I'm currently implementing backchannel logout and have started adding functionality to the work done here. Any concerns around the approach or has development stopped because of other work?
> 
> Hi, work here has stopped because the JWT implementation has some issues (including missing support for JWE and it requires some unnecessary code duplication). A friend of mine and me have started a JOSE implementation that tries to solve these problems with a better, more extensible, and powerful way including an rust only crypto backend (which solves #58). We’re aiming for our first stable release within the next two weeks (see [minkan-chat/jose#30](https://github.com/minkan-chat/jose/issues/30) and also look at the other branches). Current state is that JWS is working with a great api and we just want to implement an better way for payloads (needed because of detached content). When we’re ready, back channel tokens can easily be parsed.

Thanks @Erik1000. I've made some progress with implementation of the logout token but will look forward to having a look at your JOSE implementation.

KevinNaidoo:
> > > Hi @Erik1000 / @ramosbugs,
> > > I'd like to move this ahead, I'm currently implementing backchannel logout and have started adding functionality to the work done here. Any concerns around the approach or has development stopped because of other work?
> > 
> > 
> > Hi, work here has stopped because the JWT implementation has some issues (including missing support for JWE and it requires some unnecessary code duplication). A friend of mine and me have started a JOSE implementation that tries to solve these problems with a better, more extensible, and powerful way including an rust only crypto backend (which solves #58). We’re aiming for our first stable release within the next two weeks (see [minkan-chat/jose#30](https://github.com/minkan-chat/jose/issues/30) and also look at the other branches). Current state is that JWS is working with a great api and we just want to implement an better way for payloads (needed because of detached content). When we’re ready, back channel tokens can easily be parsed.
> 
> Thanks @Erik1000. I've made some progress with implementation of the logout token but will look forward to having a look at your JOSE implementation.

Ps., but I've seen exactly what you mean, and was why I wanted to implement the backchannel logout within the library, for JWE.

Erik1000:
> Ps., but I've seen exactly what you mean, and was why I wanted to implement the backchannel logout within the library, for JWE.

@KevinNaidoo well, JWE is quite complex and isn't implemented by `openidconnect` (and it AFAIK violates the spec because of that?). We plan to implement it, but because JWSes are used most of the time (many people don't even know that JWTs can be encrypted, almost no JWT library supports it), we want to create a first release without JWS. The groundwork for JWE has been done on our end and we have some ideas how to implement it from an API perspective.

KevinNaidoo:
> 

Feels like I'm following in your footsteps... I've been asking the question of but what if the JWT is encrypted, but many people/libraries aren't seeming to be doing that.  But it is complex, look forward to your API, makes sense to release without first.

Issue #60, opened 2022-01-16
Unresolved "env" dependencies in wasm module


coderPaddyS:
Hey there!

I a trying to use your library in a software project at our university and switched from your `oauth2` library to this one.
Sadly it is not working for me, there are unresolved dependencies which are then mapped to `env` by `wasm_bindgen`. Here is the dissassembly by wasm2wat for the relevant sections:

```wat
  (import "env" "LIMBS_equal" (func $env.LIMBS_equal (type $t6)))
  (import "env" "LIMBS_are_even" (func $env.LIMBS_are_even (type $t3)))
  (import "env" "GFp_bn_neg_inv_mod_r_u64" (func $env.GFp_bn_neg_inv_mod_r_u64 (type $t29)))
  (import "env" "LIMBS_shl_mod" (func $env.LIMBS_shl_mod (type $t4)))
  (import "env" "LIMBS_reduce_once" (func $env.LIMBS_reduce_once (type $t1)))
  (import "env" "GFp_bn_from_montgomery_in_place" (func $env.GFp_bn_from_montgomery_in_place (type $t21)))
  (import "env" "GFp_limbs_mul_add_limb" (func $env.GFp_limbs_mul_add_limb (type $t12)))
  (import "env" "LIMBS_less_than" (func $env.LIMBS_less_than (type $t6)))
  (import "env" "LIMBS_less_than_limb" (func $env.LIMBS_less_than_limb (type $t6)))
  (import "env" "GFp_memcmp" (func $env.GFp_memcmp (type $t6)))
  (import "env" "GFp_nistz256_sqr_mont" (func $env.GFp_nistz256_sqr_mont (type $t0)))
  (import "env" "GFp_nistz256_point_mul" (func $env.GFp_nistz256_point_mul (type $t4)))
  (import "env" "GFp_p256_scalar_mul_mont" (func $env.GFp_p256_scalar_mul_mont (type $t1)))
  (import "env" "GFp_p256_scalar_sqr_mont" (func $env.GFp_p256_scalar_sqr_mont (type $t0)))
  (import "env" "GFp_p256_scalar_sqr_rep_mont" (func $env.GFp_p256_scalar_sqr_rep_mont (type $t1)))
  (import "env" "GFp_nistz384_point_mul" (func $env.GFp_nistz384_point_mul (type $t4)))
  (import "env" "GFp_p384_scalar_mul_mont" (func $env.GFp_p384_scalar_mul_mont (type $t1)))
[...]

PR #72, opened 2022-04-12
fix: don't depend on C code when compiling for wasm

ctron:
The ring functionality used to check if the nonce is equal makes uses
of a memcpy function imported from the C world. However, this function
isn't available when compiling for wasm32.

This change uses a simple compare for wasm32, and keeps the current
behavior for all other targets.

closes issue #60

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -11,7 +11,6 @@ use http::method::Method;
 use http::status::StatusCode;
 use oauth2::helpers::deserialize_space_delimited_vec;
 use rand::{thread_rng, Rng};
-use ring::constant_time;
 use serde::de::DeserializeOwned;
 use serde::Serialize;
 use thiserror::Error;
@@ -917,10 +916,19 @@ new_secret_type![
         }
     }
 ];
+
+#[cfg(not(target_arch = "wasm32"))]
 impl PartialEq for Nonce {
     fn eq(&self, other: &Self) -> bool {
-        constant_time::verify_slices_are_equal(self.secret().as_bytes(), other.secret().as_bytes())
+        ring::constant_time::verify_slices_are_equal(self.secret().as_bytes(), other.secret().as_bytes())
             .is_ok()
+    }
+}
+
+#[cfg(target_arch = "wasm32")]
+impl PartialEq for Nonce {
+    fn eq(&self, other: &Self) -> bool {
+        self.0 == other.0
     }
 }
 


ctron:
Ok, looks like that was just the tip of the iceberg when pulling in the dependency. When actually using the code it get's way worse:

```
  (import "env" "GFp_bn_neg_inv_mod_r_u64" (func $GFp_bn_neg_inv_mod_r_u64 (type 70)))
  (import "env" "LIMBS_shl_mod" (func $LIMBS_shl_mod (type 15)))
  (import "env" "GFp_bn_from_montgomery_in_place" (func $GFp_bn_from_montgomery_in_place (type 25)))
  (import "env" "GFp_limbs_mul_add_limb" (func $GFp_limbs_mul_add_limb (type 16)))
  (import "env" "LIMBS_equal" (func $LIMBS_equal (type 13)))
  (import "env" "LIMBS_less_than" (func $LIMBS_less_than (type 13)))
  (import "env" "LIMBS_less_than_limb" (func $LIMBS_less_than_limb (type 13)))
  (import "env" "LIMBS_are_zero" (func $LIMBS_are_zero (type 10)))
  (import "env" "LIMBS_are_even" (func $LIMBS_are_even (type 10)))
  (import "env" "LIMB_shr" (func $LIMB_shr (type 10)))
  (import "env" "LIMBS_reduce_once" (func $LIMBS_reduce_once (type 12)))
  (import "env" "GFp_memcmp" (func $GFp_memcmp (type 13)))
  (import "env" "GFp_nistz256_point_mul" (func $GFp_nistz256_point_mul (type 15)))
  (import "env" "GFp_p256_scalar_mul_mont" (func $GFp_p256_scalar_mul_mont (type 12)))
  (import "env" "GFp_p256_scalar_sqr_mont" (func $GFp_p256_scalar_sqr_mont (type 9)))
  (import "env" "GFp_p256_scalar_sqr_rep_mont" (func $GFp_p256_scalar_sqr_rep_mont (type 12)))
  (import "env" "GFp_nistz384_point_mul" (func $GFp_nistz384_point_mul (type 15)))
  (import "env" "GFp_p384_scalar_mul_mont" (func $GFp_p384_scalar_mul_mont (type 12)))
  (import "env" "GFp_p384_elem_mul_mont" (func $GFp_p384_elem_mul_mont (type 12)))
  (import "env" "GFp_nistz256_add" (func $GFp_nistz256_add (type 12)))
  (import "env" "GFp_nistz256_mul_mont" (func $GFp_nistz256_mul_mont (type 12)))
  (import "env" "GFp_nistz256_sqr_mont" (func $GFp_nistz256_sqr_mont (type 9)))
  (import "env" "GFp_nistz256_point_add" (func $GFp_nistz256_point_add (type 12)))
  (import "env" "GFp_p384_elem_add" (func $GFp_p384_elem_add (type 12)))
  (import "env" "GFp_nistz384_point_add" (func $GFp_nistz384_point_add (type 12)))
```

ramosbugs:
Hey, thanks for looking into this. As I think you've deduced, #58 is probably the right solution here. I haven't reviewed all of the changes in that PR yet, but one comment that may apply to both PRs is that to avoid [timing side-channel attacks](https://en.wikipedia.org/wiki/Timing_attack), we'll need either a constant-time function for comparing secrets (e.g., nonces), or we'll need to first compute a secure hash (e.g., SHA-2) and compare the hashes, rather than using a simple `==` check.

sbihel:
Sorry for the delay, to answer the comment about constant-time comparison, in my PR I'm using the `subtle` crate (which is also used in the RustCrypto crates).

PR #73, opened 2022-04-29
Deps update pass

jelmansouri:
We use cargo deny to lint our dependencies on my project, sha2 seems is out of date, but I took the opportunity to update the other dev deps. Hopefully this is all still compliant with the MSRV.
Also tried to check if it affects the public interface, but hard to say.
Also untrusted doesn't seem to be used.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,6 @@ serde_derive = "1.0"
 serde_json = "1.0"
 serde_path_to_error = "0.1"
 serde-value = "0.7"
-untrusted = "0.9"
 url = { version = "2.1", features = ["serde"] }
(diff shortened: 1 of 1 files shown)


ramosbugs:
Thanks!

jelmansouri:
Thanks for accepting the PR! Can a version of this crate and (oauth2) be released if possible? part of the reason we updated is because there is a security advisory in ordered-float, which is a dependency of serde-value 0.6, that I updated to serde-value 0.7 in this PR.

ramosbugs:
This is now released in https://crates.io/crates/openidconnect/2.3.1

jelmansouri:
Thanks again ! Have a nice weekend !

PR #84, opened 2022-09-05
introduce `StringOrBool` to workaround illegal implementation

Justice4Joffrey:
This is a kludge I'm using to work with Apple OAuth. They send string bools for these fields, sadly.

It's not really a merge-worthy fix, but if you can suggest a more appropriate way of integrating this (I'm sure it's a large use-case for your consumers of this library), I'm happy to throw it together.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ ureq = ["oauth2/ureq"]
 native-tls = ["oauth2/native-tls"]
 rustls-tls = ["oauth2/rustls-tls"]
 accept-rfc3339-timestamps = []
+accept-string-booleans = []
 nightly = []
 
 [dependencies]
--- a/src/claims.rs
+++ b/src/claims.rs
@@ -9,7 +9,7 @@ use serde::{Serialize, Serializer};
 
 use crate::helpers::FlattenFilter;
 use crate::types::helpers::{split_language_tag_key, timestamp_to_utc, utc_to_seconds};
-use crate::types::{LocalizedClaim, Timestamp};
+use crate::types::{Boolean, LocalizedClaim, Timestamp};
 use crate::{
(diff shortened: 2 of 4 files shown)


Justice4Joffrey:
This is a kludge I'm using to work with Apple OAuth. They send string bools for these fields, sadly.

It's not really a merge-worthy fix, but if you can suggest a more appropriate way of integrating this (I'm sure it's a large use-case for your consumers of this), I'm happy to throw it together.

ramosbugs:
hey, thanks for the PR!

In order to avoid introducing a breaking change, I'd suggest only changing the `Deserialize` implementation, rather than the actual field type. Also, for this crate to remain standards-compliant by default, relaxations like this should be behind a non-default feature flag.

For this change, let's follow the example of the `accept-rfc3339-timestamps` feature flag: https://github.com/ramosbugs/openidconnect-rs/blob/769f796ccf8b45bacf7bdbfb9dc8184d2a590b66/src/types.rs#L1192-L1216

Internally, it uses a `Timestamp` type that deserializes from either an integer or a string depending on whether the feature flag is enabled: https://github.com/ramosbugs/openidconnect-rs/blob/769f796ccf8b45bacf7bdbfb9dc8184d2a590b66/src/types.rs#L1004-L1010

In this case, however, we should only need to specify a `deserialize_with` annotation since the underlying type (`bool`) already matches the `Serialize` implementation we want (vs. needing to convert a `chrono::DateTime` back to seconds).

Justice4Joffrey:
Thanks for the pointers. Might be some stylistic things to clear up, but think this is largely what you were suggesting.

Let me know!

ramosbugs:
This has now been released in [2.4.0](https://crates.io/crates/openidconnect/2.4.0)!

Issue #86, opened 2022-09-20
Need help understanding what's wrong with my id token


DanielMalmgren:
Hi.
I hope you can give me some help. I'm using a Rust program built on openidconnect-rs and it doesn't accept the id tokens sent by my idp and I'm trying to understand why. Unfortunately I have no experience in Rust and I have big problems understanding the syntax. It makes a call to the claims() method in IdToken and get a ClaimsVerificationError back. By adding the following code to the programs AppError class:
```
impl From<openidconnect::ClaimsVerificationError> for AppError {
    fn from(error: openidconnect::ClaimsVerificationError) -> Self {
        info!("Claims error: {}", error);
        AppError::InvalidIdTokenNonce
    }
}
```
...i got closer, now I get "Claims error: Signature verification failed" in my log. However I seem to fail in getting to know exactly why it doesn't like the signature. If I understand things correctly [here](https://github.com/ramosbugs/openidconnect-rs/blob/main/src/verification.rs#L78-L116) I should be able to get info about what exactly it thinks is the problem, but I don't know how.

I've spent far too many hours on this, can someone please help me? I won't hesitate donating some bucks if I solve this...

ramosbugs:
Hi Daniel,

Welcome to the Rust community! The error types returned by this crate implement cause chaining similar to errors in Java. Specifically, they implement the [`std::error::Error`](https://doc.rust-lang.org/nightly/std/error/trait.Error.html) trait (see [`ClaimsVerificationError`](https://docs.rs/openidconnect/2.3.2/openidconnect/enum.ClaimsVerificationError.html#impl-Error) docs), which provides a [`source`](https://doc.rust-lang.org/nightly/std/error/trait.Error.html#method.source) method for accessing the underlying error. You can walk that chain recursively and print out the errors to see what happened.

Hope that helps!

DanielMalmgren:
Ah! After some experimenting, I found out that error.source() gave me exactly what I want. So now I know what it's complaining about, but not how to solve it. The error it's giving me is DisallowedAlg("algorithm `RS512` is not one of: RS256") and if I understand correctly that means I have to add RS512 to "id_token_signed_response_alg", but where do I do that? Seems to have something with dynamic client registration to do, but it's nothing I have ever used...

ramosbugs:
[...]

PR #87, opened 2022-09-21
Added support for discovering supported id token signing algorithms

MarcelCoding:
#86
Mentioned here: https://github.com/ramosbugs/openidconnect-rs/issues/86#issuecomment-1252061758

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -740,6 +740,7 @@ pub enum AuthenticationFlow<RT: ResponseType> {
 /// #     Some(TokenUrl::new("https://example/token".to_string()).unwrap()),
 /// #     None,
 /// #     JsonWebKeySet::default(),
+/// #     vec![],
 /// # )
 /// # .set_revocation_uri(RevocationUrl::new("https://revocation/url".to_string()).unwrap());
 /// #
@@ -800,6 +801,7 @@ where
     issuer: IssuerUrl,
     userinfo_endpoint: Option<UserInfoUrl>,
     jwks: JsonWebKeySet<JS, JT, JU, K>,
+    id_token_signing_algs: Vec<JS>,
     use_openid_scope: bool,
     _phantom: PhantomData<(AC, AD, GC, JE, P)>,
 }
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -833,6 +835,7 @@ where
         token_url: Option<TokenUrl>,
         userinfo_endpoint: Option<UserInfoUrl>,
         jwks: JsonWebKeySet<JS, JT, JU, K>,
+        id_token_signing_algs: Vec<JS>,
```
ramosbugs, line 838: I'd prefer not to introduce this as a breaking change to the public API. To avoid this, the `from_provider_metadata` function could either construct `Self` directly or this can be factored into a separate private `new_impl` method (either is fine).

src/lib.rs
```diff
@@ -955,12 +962,14 @@ where
                 client_secret.clone(),
                 self.issuer.clone(),
                 self.jwks.clone(),
+                self.id_token_signing_algs.clone(),
```
ramosbugs, line 965: We can avoid breaking changes here by just calling `set_allowed_algs` on the new verifier before returning it

ramosbugs (commented):
Thanks for the PR!

src/lib.rs
```diff
@@ -846,6 +847,40 @@ where
...
+           id_token_signing_algs: vec![],
+            use_openid_scope: true,
+            _phantom: PhantomData,
+        }
+    }
+
+    ///
+    /// Initializes an OpenID Connect client with predefined algorithms.
+    /// <br>
+    /// **LEAVE THIS PRIVATE UNTIL ALL ALGORITHMS FROM DISCOVERY ARE IMPLEMENTED**
```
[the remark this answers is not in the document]
MarcelCoding: Yes that seams reasonable. I'll going to change the comment to something hinting if configuration is needed you should do it directly on the ID token verifier.

MarcelCoding:
@ramosbugs Did I make this a draft? Is there is there any reason? I don't see any. If you don't have anything left, this is ready for another review.

ramosbugs:
> @ramosbugs Did I make this a draft?

I think so :-)

I'll take another look this afternoon

MarcelCoding:
Thanks!

ramosbugs:
This is now released in [2.5.0](https://crates.io/crates/openidconnect/2.5.0)!

Issue #77, opened 2022-07-04
Error validating issuer URI


pinpox:
I'm not sure what is happening here, as the "expected" value in the error I'm getting is the same as the one I provided.

This is my code snippet:
```rust
async fn get_client_from_sso_config(sso_config: &SsoConfig) -> Result<CoreClient, &'static str> {
    let redirect = sso_config.callback_path.to_string();
    let client_id = ClientId::new(sso_config.client_id.as_ref().unwrap().to_string());
    let client_secret = ClientSecret::new(sso_config.client_secret.as_ref().unwrap().to_string());
    let issuer_url =
        IssuerUrl::new(sso_config.authority.as_ref().unwrap().to_string()).or(Err("invalid issuer URL"))?;
    let test = CoreProviderMetadata::discover_async(issuer_url, async_http_client).await;

    println!("TEST43 {:?}", test);

    let provider_metadata = match test {
        Ok(metadata) => {
            println!("TEST46");
            metadata
        }
        Err(_err) => {
            println!("TEST47");
            return Err("Failed to discover OpenID provider");
        }
    };
// other code
}
```
When executed I get this error:
```
TEST43 Err(Validation("unexpected issuer URI `https://git.lounge.rocks/` (expected `https://git.lounge.rocks/`)"))
TEST47
[2022-07-04 15:12:35.409][vaultwarden::api::identity][ERROR] Unable to find client from identifier
```
Notice the line  `"unexpected issuer URI https://git.lounge.rocks/ (expected https://git.lounge.rocks/)"` Not sure if this is a bug on my side or a wrong error message, but I can't figure out what is wrong here.

ramosbugs:
Hey, thanks for filing this issue. I pushed a patch in the new 2.3.2 release to make the error message clearer by printing the raw issuer URL that's involved in the verification check rather than the canonicalized `Url` object. I suspect the issue is that one of the issuer URLs has a trailing slash and the other does not, but hopefully updating to the latest version will reveal the underlying issue.

pinpox:
It was indeed a trailing slash! Should the validation should be changed to remove trailing slashes or is this something outside the scope of openidconnect-rs?

ramosbugs:
I think the current validation should work; clients just need to be sure that the issuer URL they provide exactly matches the one that the OIDC provider returns in its discovery metadata and JWTs. The actual discovery fetch should work whether or not the trailing slash is included, so it's just a matter of satisfying the validation.

[...]

PR #88, opened 2022-09-29
Compare equality by URL rather than by string

bstrie:
There exist distinct strings that will parse to equivalent URLs. For example:

```rust
use url::Url;

fn main() {
    let x: Url = "http://example.com".parse().unwrap();
    let y: Url = "http://example.com/".parse().unwrap();
    assert_eq!(x, y); // passes
}
```

Currently, `IssuerUrl` (and every other URL type) compares for equality via the string it was parsed from, rather than the URL that it was parsed into. However, this is problematic in at least one context: when discovering provider metadata, the discovered URL is validated against the URL that was passed in. The check in question is here: https://github.com/ramosbugs/openidconnect-rs/blob/2f79ac2ec699eff2f17a2b5c50290782623d8a9f/src/discovery.rs#L404 . In this case, we have no control over the response from the server, so it's a bit frustrating when the call fails for the simple reason of including/omitting a trailing slash. It also risks accidental breakage if a server makes a harmless-seeming change to include or omit the trailing slash, for whatever reason.

(To be clear, this failure mode was much *worse* prior to https://github.com/ramosbugs/openidconnect-rs/commit/87cf0d79e4b4732e43e034f5104e99b039d1f671 .)

---

If you're wary of this approach, I can propose two alternatives:

1. If you're worried about two IssuerUrls comparing equal while containing different strings, then when the IssuerUrl is constructed here: https://github.com/ramosbugs/openidconnect-rs/blob/2f79ac2ec699eff2f17a2b5c50290782623d8a9f/src/macros.rs#L317 we could canonicalize the stored String by converting the parsed URL back into a string and storing that, rather than storing the input directly.
2. Alternatively, if you want the IssuerUrl behavior to remain the same, then this could also be fixed by changing the ProviderMetadata validation (linked above) to explicitly compare via the stored URL, rather than comparing via the equality of the IssuerUrl itself.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -390,7 +390,7 @@ macro_rules! new_url_type {
         }
         impl PartialEq for $name {
             fn eq(&self, other: &$name) -> bool {
-                self.1 == other.1
+                self.0 == other.0
             }
         }
         impl Eq for $name {}
--- a/src/types.rs
+++ b/src/types.rs
@@ -1399,6 +1399,14 @@ mod tests {
         );
     }
 
+    #[test]
+    fn issuer_url_equality() {
+        let issuer_url_1 = IssuerUrl::new("http://example.com".to_string()).unwrap();
+        // Note the deliberate trailing slash in the following URL
+        let issuer_url_2 = IssuerUrl::new("http://example.com/".to_string()).unwrap();
+        assert_eq!(issuer_url_1, issuer_url_2);
+    }
+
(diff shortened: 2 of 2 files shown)


ramosbugs:
Unfortunately, the current behavior is required by the spec. See https://github.com/ramosbugs/openidconnect-rs/issues/77#issuecomment-1175397724.

PR #92, opened 2022-10-21
Core JWK use type to ignore  unknown use types

### LINKED ISSUE #91: It's not Obvious how to Specialize ProviderMetadata
I've got an OIDC provider, which in the discovery document may specify keys with other uses than `sig` or `enc`. This means that while trying to load Discovery data, I get the error:

```
keys[0].use: unknown variant `saml_ca`, expected `sig` or `enc` at line 1 column 25
unknown variant `saml_ca`, expected `sig` or `enc` at line 1 column 25', src\main.rs:73:48
```

I assume that the intention here is that I can specify my own `ProviderMetadata` instead of using `CoreProviderMetadata`, setting `CoreJsonWebKeyUse` to my own enum. However, this appears to be a futile exercise as best I can tell.

To override `CoreJsonWebKeyUse`, I also need to have to replace `CoreJsonWebKey`, which needs to implement `JsonWebKey`. Unfortunately, the majority of the code used to implement this within this code base is private. Which means basically I just have to rewrite the entire crate to do this.

Hopefully I've simply misunderstood something here and this is much easier than I'm making it for myself?

Alexei-Barnes:
This is to comply with RFC 7517, Section 4: "Additional members can be present in the JWK; if not understood by implementations encountering them, they MUST be ignored"

Some providers have keys with uses that don't apply to our use case, this change enables us to ignore them gracefully.

This should fix #91 for the simple case.

### DIFF
--- a/src/core/jwk.rs
+++ b/src/core/jwk.rs
@@ -497,6 +497,11 @@ pub enum CoreJsonWebKeyUse {
     ///
     #[serde(rename = "enc")]
     Encryption,
+    ///
+    /// Key has some other use we don't know how to make use of.
+    ///
+    #[serde(other)]
+    Unknown,
 }
 impl JsonWebKeyUse for CoreJsonWebKeyUse {
     fn allows_signature(&self) -> bool {


ramosbugs:
Thanks! I pushed an alternative in c25eb06ba7e64e14cb94d34a564d77a4ff4e9d1b that captures the string value.

PR #93, opened 2022-10-27
Added Send to CoreRsaPrivateSigningKey.rng

Lzok:
Hello! How are you?

It looks like the type that holds RSA keys from the `openidconnectcrate` (`CoreRsaPrivateSigningKey`) is not `Send` because it holds a reference to a `ring::rand:SecureRandom + 'static`. This seems a bit odd to me in real use cases (e.g. backends with multiple threads), since it doesn’t look like a key should be something `!Send`.

One question I ask to myself is why can’t I share a single key among several worker threads signing tokens? Do you have any concern with this that I am not being aware of? 

**Important Note**
This PR does not break any public API, that's why I am pull requesting it directly.

I would like to hear your thoughts on this and how we can improve the solution together.

Thank you!

### DIFF
--- a/src/core/jwk.rs
+++ b/src/core/jwk.rs
@@ -323,7 +323,7 @@ const RSA_FOOTER: &str = "-----END RSA PRIVATE KEY-----";
 ///
 pub struct CoreRsaPrivateSigningKey {
     key_pair: ring_signature::RsaKeyPair,
-    rng: Box<dyn rand::SecureRandom>,
+    rng: Box<dyn rand::SecureRandom + Send>,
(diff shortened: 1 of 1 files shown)


ramosbugs:
Thanks for the PR!

> Do you have any concern with this that I am not being aware of?

Nope, just an oversight. Will merge once CI passes. I agree that this doesn't appear to be a breaking change since it only affects internal APIs.

Lzok:
@ramosbugs 
Thanks a lot!

ramosbugs:
This has now been released in [2.4.0](https://crates.io/crates/openidconnect/2.4.0)!

PR #94, opened 2022-11-10
feat: Abstract crypto backends

ctron:
This is a draft/preview/idea on how to abstract the crypto backend.

This wouldn't rely on any `feature` flags, but would allow the user to choose. Which might make sense for a library.

The follows the idea of e.g.: https://github.com/sfackler/rust-postgres/blob/71668b7f460955750bcfa23ffb3efb022915a6d0/postgres/src/lib.rs#L9

With the difference that you need to specify only a type, not an instance. And that there is a default type.

I am not 100% this idea works out, but I wanted to have some feedback before continuing the work.

### DIFF
--- a/src/core/crypto/mod.rs
+++ b/src/core/crypto/mod.rs
@@ -1,0 +1,28 @@
+use crate::SigningError;
+use ::ring::rand::SecureRandom;
+use ::ring::signature as ring_signature;
+
+mod ring;
+
+/// Cryptography backend
+pub trait Backend {
+    fn sign_rsa(
+        key: &ring_signature::RsaKeyPair,
+        padding_alg: &'static dyn ring_signature::RsaEncoding,
+        rng: &dyn SecureRandom,
+        msg: &[u8],
+    ) -> Result<Vec<u8>, SigningError>;
+}
+
+pub struct Ring;
+
+impl Backend for Ring {
+    fn sign_rsa(
+        key: &ring_signature::RsaKeyPair,
+        padding_alg: &'static dyn ring_signature::RsaEncoding,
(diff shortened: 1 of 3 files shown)


ctron:
Currently it is expected that the code will not compile.

sbihel:
Hi, just jumping in to say that if you want to abstract the crypto backend and make it configurable, it's probably better to use https://crates.io/crates/signature or https://crates.io/crates/signatory. 

Sorry for leaving #58 untouched, I've been busy.

ctron:
> Hi, just jumping in to say that if you want to abstract the crypto backend and make it configurable, it's probably better to use https://crates.io/crates/signature or https://crates.io/crates/signatory.
> 
> Sorry for leaving #58 untouched, I've been busy.

No worries. We all got lives :grinning:  … I wanted to split this up into smaller bits. So, the next step (this one) would be to create an abstraction layer. To confine the use of `ring` into one module, and allowing for an additional implementation. Of course, it would make sense to use the crates you mentioned.

ramosbugs:
if we can achieve feature parity (including platform support being a superset of ring's), I'm happy to migrate entirely off of ring and avoid the complexity of an abstraction layer. is there a use case for retaining ring support?

ctron:
> if we can achieve feature parity (including platform support being a superset of ring's), I'm happy to migrate entirely off of ring and avoid the complexity of an abstraction layer. is there a use case for retaining ring support?

Not from my side, I assumed you wanted to keep it. Yes, that might make things a lot easier.

In this case, just close this PR, and I will continue with PRs, slowly kicking out ring, until it is gone.

PR #104, opened 2023-03-09
fix: don't enforce jwk signing algorithm if empty

nshaaban-cPacket:
Makes the CoreClient::new function usable for claim verification again, as it was in 2.4, while still enforcing IF we have enough context to do so, improving usability of the library.

When attempting to verify a claim using CoreClient::new (as done in 2.4) in version 2.5, the error `Signature verification failed\\n\\nCaused by:\\n    Disallowed signature algorithm: algorithm `RS256` is not one of: \"` is returned (due to Client::new setting an empty array for supported algorithms).

## Context:
OIDC provider: auth0 or google
version: 2.5

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -800,7 +800,7 @@ where
     issuer: IssuerUrl,
     userinfo_endpoint: Option<UserInfoUrl>,
     jwks: JsonWebKeySet<JS, JT, JU, K>,
-    id_token_signing_algs: Vec<JS>,
+    id_token_signing_algs: Option<Vec<JS>>,
     use_openid_scope: bool,
     _phantom: PhantomData<(AC, AD, GC, JE, P)>,
 }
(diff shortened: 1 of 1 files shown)


ramosbugs:
Thanks for the PR! This bug fix makes sense. I'll backport it to the `support/2.x` branch and cut a 2.5.1 release since 3.0 isn't ready yet.

src/lib.rs
```diff
@@ -825,8 +825,9 @@ where
 {
     ///
     /// Initializes an OpenID Connect client.
-    /// If you need to configure the algorithms used for signing, ...,
-    /// do this directly on the respected components. (e.g. IdTokenVerifier)
+    /// WARNING: does not restrict id token signing algorithms,
```
ramosbugs, line 828: I think we can remove this comment because not calling `set_allowed_algs` below will default to `[RS256]` as indicated in the [spec](https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation).

ramosbugs (commented):
oh, this *is* on the `support/2.x` branch. in that case I'll forward port it to `main` after merging :-)

src/lib.rs
nshaaban-cPacket: Sounds good! Done.

ramosbugs: that happens here, reference: https://github.com/ramosbugs/openidconnect-rs/blob/a0434bb7a559e09b1de8dfbee978af7c37439eda/src/verification.rs#L148

ramosbugs:
This is now released in [2.5.1](https://crates.io/crates/openidconnect/2.5.1)

Issue #79, opened 2022-07-12
Provide AuthorizationRequest::add_scope variant that does not consume scope


m-thielen:
[Disclaimer: I'm pretty new to Rust, I sincerely hope this all makes sense :-)]

I'm using openidconnect-rs in a project where the OpenID Connect scope is configurable; it is read from a config file or the environment as a space separated string. In order to add all scopes to the authorization URL, I have to call `add_scope` repeatedly and since `add_scope` consumes the scope, I have to clone it. This seems rather inefficient to me, since this has to be done for each session that is created.

So I think a variant that accepts an `Rc<Scope>` or `Arc<Scope>` would be helpful.

ilmari-h:
I'm running into the exact same problem.
This repo uses the [oauth2 crate ](https://docs.rs/oauth2/latest/oauth2/index.html) and calls its equivalent method, `oauth2::AuthorizationRequest::add_scope`.

In the oauth2 crate, there is also a function [`oauth2::AuthorizationRequest::add_scopes`](https://docs.rs/oauth2/latest/oauth2/struct.AuthorizationRequest.html#method.add_scopes) which takes an iterator of scopes, allowing you to pass in multiple in one call. I think this would be a pretty simple improvement, probably just requires adding the equivalent function here.

ramosbugs:
thanks @ilmari-h for providing that PR. I think it's a worthwhile improvement to add to this crate, but I don't think it addresses the issue OP posted above, which is about shared references to scopes. The new function still takes ownership of the scopes and would require cloning the input if it needs to be reused for subsequent requests.

As for OP's issue with cloning scopes, I think this is a level of micro-optimization that isn't worth the complexity it would add. Reference-counted types also add an additional layer of indirection/dereference that comes with its own performance overhead. If we were talking about large data structures or a tight loop where we'd be cloning many times in a CPU-bound part of the code, it would be worth it. But, the time here is dominated by network requests, and it's not worth focusing on a few short strings being cloned once per request.

PR #105, opened 2023-03-21
add AuthorizationRequest::add_scopes

ilmari-h:
Added `AuthorizationRequest::add_scopes`. 
This is an equivalent to the underlying [`oauth2::AuthorizationRequest::add_scopes`](https://docs.rs/oauth2/latest/oauth2/struct.AuthorizationRequest.html#method.add_scopes)  method, that instead takes an iterator of Scopes as a parameter.

This enables adding a list of scopes in one go, and should solve the use case described in issue https://github.com/ramosbugs/openidconnect-rs/issues/79

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1211,7 +1211,7 @@ where
     ///
     pub fn add_scopes<I>(mut self, scopes: I) -> Self
     where
-        I : IntoIterator<Item = Scope>,
+        I: IntoIterator<Item = Scope>,
     {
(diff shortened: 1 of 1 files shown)


ramosbugs:
Thanks for the PR! Please just fix the cargo fmt issue (pinned to 1.57.0 in case it's version-dependent), and then I'll merge it once CI passes.

ramosbugs:
This is now released in [3.0.0](https://crates.io/crates/openidconnect/3.0.0)!

PR #107, opened 2023-04-08
Remove unnecessary 'static bound from `ProviderMetadata::discover_async`

ATiltedTree:
Allows for `http_client` to be a closure which captures values by non-static reference.


ramosbugs:
Thanks for the PR! I'm not sure why I had this bound in the first place since it's not present in the other HTTP client callbacks.

ramosbugs:
This is now released in [3.0.0](https://crates.io/crates/openidconnect/3.0.0)!

Issue #111, opened 2023-05-19
Missing method to retrieve end_session_endpoint or logout


jsimonrichard:
It would be nice to be able to log out using this framework. OpenID Connect requires there to be an end_session_endpoint URL, but as far as I can tell this library is missing a method to call or get that URL.

Is this the same as the revocation URL in the Google example or is this something different? It's present on the JSON discovery document at `.well-known/openid-configuration`.

ramosbugs:
Hi @jsimonrichard,

This crate doesn't currently support the [OpenID Connect RP-Initiated Logout 1.0](https://openid.net/specs/openid-connect-rpinitiated-1_0.html) specification, and this field is not part of any of the supported standards listed on the [README](https://github.com/ramosbugs/openidconnect-rs#standards).

I would be open to a contribution that adds complete support for this standard, which I think should include:
- The `end_session_endpoint` in the OIDC Discovery metadata. Since this field isn't part of the OIDC Core standard, I wouldn't add it to `ProviderMetadata`. Rather, let's leverage `ProviderMetadata`'s `A: AdditionalProviderMetadata` type argument. We can do something along the lines of the following:
```rust
#[non_exhaustive]
#[derive(Clone, Debug, Deserialize, Serialize)]
pub struct LogoutProviderMetadata<A>
where A: AdditionalProviderMetadata {
    // Optional even though it's required by the spec so that clients can use the
    // field to determine whether the OP supports RP-initiated logout.
    pub end_session_endpoint: Option<EndSessionUrl>,
    #[serde(bound = "A: AdditionalProviderMetadata", flatten)]
    pub additional_metadata: A,
}
impl<A> AdditionalProviderMetadata for LogoutProviderMetadata<A>
where A: AdditionalProviderMetadata {}

pub type ProviderMetadataWithLogout = ProviderMetadata<
    LogoutProviderMetadata<EmptyAdditionalProviderMetadata>,
    CoreAuthDisplay,
    CoreClientAuthMethod,
    CoreClaimName,
    CoreClaimType,
    CoreGrantType,
    CoreJweContentEncryptionAlgorithm,
    CoreJweKeyManagementAlgorithm,
    CoreJwsSigningAlgorithm,
    CoreJsonWebKeyType,
    CoreJsonWebKeyUse,
    CoreJsonWebKey,
    CoreResponseMode,
    CoreResponseType,
    CoreSubjectIdentifierType,
>;
```
This would allow client code to be written as follows:
[...]

PR #112, opened 2023-05-19
Add end_session_endpoint

jsimonrichard:
A (maybe temporary) solution to #111.

Simply creates an associated type and adds the field to ProviderMetadata.


jsimonrichard:
I finished a draft for supporting RP-Initiated logout, as recommended by @ramosbugs in #111.

I haven't added any test coverage yet; I wanted to wait until I got some feedback since I'm not very familiar with OpenID.

src/lib.rs
```diff
@@ -1475,6 +1478,122 @@ where
...
+    pub fn set_client_id(mut self, client_id: ClientId) -> Self {
+        self.parameters.client_id = Some(client_id);
+        self
+    }
+
+    ///
+    /// Provides the OpenID Connect Provider with a URI to redirect to after
+    /// the logout has been performed.
+    ///
+    pub fn set_redirect_uri(mut self, redirect_uri: RedirectUrl) -> Self {
```
ramosbugs, line 1562: Let's call this `set_post_logout_redirect_uri`. Also, the `RedirectUrl` type is intended to refer to the authorization endpoint's redirect URI. Since these redirect URLs serve different purposes and are used for different endpoints, let's define a separate `PostLogoutRedirectUrl` type.

src/lib.rs
```diff
@@ -1475,6 +1478,122 @@ where
...
+        self.parameters.post_logout_redirect_uri = Some(redirect_uri);
+        self
+    }
+
+    ///
+    /// Specify an opaque value that the OpenID Connect Provider should pass back
+    /// to your application using the state parameter when redirecting to redirect_uri.
```
ramosbugs, line 1569: `post_logout_redirect_uri`

src/lib.rs
```diff
@@ -1475,6 +1478,122 @@ where
...
+    ///
+    pub fn add_ui_locale(mut self, ui_locale: LanguageTag) -> Self {
+        self.parameters.ui_locales.push(ui_locale);
+        self
+    }
+
+    ///
+    /// Returns the full logout URL
+    ///
+    pub fn url(self) -> Url {
```
ramosbugs, line 1590: the spec is a little ambiguous, but it says the following:
> OpenID Providers MUST support the use of the HTTP GET and POST methods defined in [RFC 7231](https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RFC7231) [RFC7231] at the Logout Endpoint. RPs MAY use the HTTP GET or POST methods to send the logout request to the OP. If using the HTTP GET method, the request parameters are serialized using URI Query String Serialization. If using the HTTP POST method, the request parameters are serialized using Form Serialization.

What I *think* this is saying (unfortunately, there are no clarifying examples), is that when using a `GET` request, the request parameters are added to the query string, and when using a `POST` request, they're part of the `application/x-www-form-urlencoded` encoded request body.

Since this function is adding the parameters to the query string, let's rename it to `http_get_url` or something. Also, the documentation for `serde_urlencoded` says that it produces an `application/x-www-form-urlencoded` string, which is the encoding for HTML forms submitted in HTTP POST bodies. I'm not certain that this is the same encoding as for query strings in every edge case. To avoid any potential encoding issues, let's build the query string the same way that the oauth2 crate does: https://github.com/ramosbugs/oauth2-rs/blob/6b21af5c04a3c03d2dfae4c5dd7199111feae778/src/lib.rs#L1139-L1165. Be sure to omit any `None` values (the current serde serialization includes them I think).

We probably don't need to add any sort of support for `POST` requests since both `GET` and `POST` requests to the logout endpoint need to happen as a browser navigation, and I think (?) the only way to make a `POST` request like that is to submit an actual HTML form element. This isn't a server-to-server endpoint, so we wouldn't want to add a method to make the POST request ourselves, which would be missing any relevant end-user browser cookies required by the OP, and which wouldn't redirect the user to the `post_logout_redirect_uri`.

src/lib.rs
```diff
@@ -1475,6 +1478,122 @@ where
...
+        let mut url = self.end_session_endpoint.url().to_owned();
+        url.set_query(serde_urlencoded::to_string(self.parameters).ok().as_deref());
```
ramosbugs, line 1592: I missed this yesterday, but I don't think we should swallow the serialization error if there is one. hopefully getting rid of `serde_urlencoded` and following `oauth2`'s approach of using `query_pairs_mut()` directly will allow us to have an infallible function, though

src/logout.rs
```diff
@@ -0,0 +1,189 @@
...
+    ///
+    /// Returns the full logout URL. In order to logout, a GET request should be made to this URL
+    /// by the client's browser.
+    ///
+    pub fn http_get_url(self) -> Url {
+        let mut url = self.end_session_endpoint.url().to_owned();
+        {
+            let mut query = url.query_pairs_mut();
+
+            macro_rules! add_pair {
```
ramosbugs, line 173: nice

src/logout.rs
```diff
@@ -0,0 +1,189 @@
...
+                    }
+                };
+            }
+
+            add_pair!(id_token_hint, id_token_hint.as_str());
+            add_pair!(logout_hint, logout_hint.secret());
+            add_pair!(client_id, client_id.as_str());
+            add_pair!(post_logout_redirect_uri, post_logout_redirect_uri.as_str());
+            add_pair!(state, state.secret());
+        }
```
ramosbugs, line 186: this is missing `ui_locales` I think

ramosbugs (commented):
Thanks for making these changes!

Just needs some test coverage, and then it should be good to merge. I'd suggest adding the following test cases to `logout.rs`:
- `ProviderMetadataWithLogout` deserialization (see [example](https://github.com/ramosbugs/openidconnect-rs/blob/1a986d7396516daa8c5f1b0890b2c9c7ca197353/src/discovery.rs#L1200))
- Minimal `http_get_url` with no optional parameters set (see [example](https://github.com/ramosbugs/openidconnect-rs/blob/1a986d7396516daa8c5f1b0890b2c9c7ca197353/src/lib.rs#L1507))
- Full `http_get_url` with all parameters set (including multiple `ui_locales`) (see [example](https://github.com/ramosbugs/openidconnect-rs/blob/1a986d7396516daa8c5f1b0890b2c9c7ca197353/src/lib.rs#L1567))

I think it's also worth adding OpenID Connect RP-Initiated Logout 1.0 to the list of supported standards in the README.

jsimonrichard:
>     * `ProviderMetadataWithLogout` deserialization (see [example](https://github.com/ramosbugs/openidconnect-rs/blob/1a986d7396516daa8c5f1b0890b2c9c7ca197353/src/discovery.rs#L1200))

It looks like I could just reuse the json from the example you provided. Should I pull the json out into a constant/static variable and then make the tests module public? Copy and paste?

ramosbugs:
> > * `ProviderMetadataWithLogout` deserialization (see [example](https://github.com/ramosbugs/openidconnect-rs/blob/1a986d7396516daa8c5f1b0890b2c9c7ca197353/src/discovery.rs#L1200))
> 
> It looks like I could just reuse the json from the example you provided. Should I pull the json out into a constant/static variable and then make the tests module public? Copy and paste?

I think I would just copy-paste the JSON, and you can probably simplify it since the main thing we care about for this test case is the new field.

jsimonrichard:
Alright, I think that should be it

ramosbugs:
thanks! I'll cut a release shortly

ramosbugs:
This is now released in [3.1.0](https://crates.io/crates/openidconnect/3.1.0)!

jsimonrichard:
Aah, I forgot to export `PostLogoutRedirectUrl` as a public type.

ramosbugs:
I just pushed a fix in [3.1.1](https://crates.io/crates/openidconnect/3.1.1)

Issue #113, opened 2023-05-26
Implementing device authorisation flow


ljyanesm:
Hi,

I've been using this crate for implementing oidc. But, have found a small difficulty with implementing the `device grant authorization flow` (https://developer.okta.com/docs/guides/device-authorization-grant/main/). The main difficult I am bumping into is that although one can extend the `ProviderMetadata` type to include some additional fields.

```rust
// Teach openidconnect-rs about the device authorization url.
#[derive(Clone, Debug, Deserialize, Serialize)]
struct DeviceEndpointProviderMetadata {
    device_authorization_endpoint: String,
}
impl AdditionalProviderMetadata for DeviceEndpointProviderMetadata {}
type DeviceAuthorizationProviderMetadata = ProviderMetadata<
    DeviceEndpointProviderMetadata,
    CoreAuthDisplay,
    CoreClientAuthMethod,
    CoreClaimName,
    CoreClaimType,
    CoreGrantType,
    CoreJweContentEncryptionAlgorithm,
    CoreJweKeyManagementAlgorithm,
    CoreJwsSigningAlgorithm,
    CoreJsonWebKeyType,
    CoreJsonWebKeyUse,
    CoreJsonWebKey,
    CoreResponseMode,
    CoreResponseType,
    CoreSubjectIdentifierType,
>;
```

Having the `device_authorization_endpoint` in the `ProviderMetadata` object is not enough to populate the underlaying `oauth2::Client` in the `CoreClient` as the internal `device_authorization_url: Option<DeviceAuthorizationUrl>` field is not reachable from outside the `openidconnect-rs` crate. 

Given there's neither a method to populate the `device_authorization_url` in the `CoreClient` nor some way (that I know of, if there's one please let me know) to extend the `CoreClient`, implementing this flow becomes a bit tricky. 

The current only option I think I have is making a full copy of the `CoreClient` in my crate to add some functionality to populate the `device_authorization_url` in the `oauth2::Client` in the `new` method or `from_provider_metadata` and add a couple of methods to expose it and prepare requests against it, but that would be mostly repeating the same code in the existing `CoreClient`.

Maybe I am holding it wrong?

Apologies if this is the case. Any guidance you could possibly share for implementing the `device grant authorization flow` would be appreciated.

Thank you very much for this crate,
Luis

ramosbugs:
Hi @ljyanesm,

[...]

PR #114, opened 2023-05-30
Implements the device authorisation flow

ljyanesm:
* Adds support for OAuth 2.0 Device Authorization Grant
* Updates `core/mod.rs` to include the new `CoreDeviceAuthorizationResponse` type
* Updates `lib.rs` to provide access to `DeviceAuthorizationUrl` and `DeviceAuthorizationRequest` types
* Adds `set_device_authorization_uri` method to the `Client` API to set the URL for contacting the device authorization endpoint
* Implements an example that showcases the process of using the device grant flow to obtain an ID token from the Okta provider. The example demonstrates device code exchange for an access token.

Solves #113

### DIFF
--- a/examples/okta_device_grant.rs
+++ b/examples/okta_device_grant.rs
@@ -1,0 +1,135 @@
+//!
+//! This example showcases the process of using the device grant flow to obtain an ID token from the
+//! [Okta](https://developer.okta.com/docs/guides/device-authorization-grant/main/#request-the-device-verification-code)
+//! provider.
+//!
+//! Before running it, you'll need to generate your own
+//! [Okta Server](https://developer.okta.com/signup/).
+//!
+//! In order to run the example call:
+//!
+//! ```sh
+//! CLIENT_ID=xxx CLIENT_SECRET=yyy ISSUER_URL=zzz cargo run --example okta_device_grant
+//! ```
+//!
+//! ...and follow the instructions.
+//!
+
+use openidconnect::core::{
+    CoreAuthDisplay, CoreClaimName, CoreClaimType, CoreClient, CoreClientAuthMethod,
+    CoreDeviceAuthorizationResponse, CoreDeviceAuthorizationUrl, CoreGrantType, CoreJsonWebKey,
+    CoreJsonWebKeyType, CoreJsonWebKeyUse, CoreJweContentEncryptionAlgorithm,
+    CoreJweKeyManagementAlgorithm, CoreJwsSigningAlgorithm, CoreResponseMode, CoreResponseType,
+    CoreSubjectIdentifierType,
+};
+use openidconnect::{
+    AdditionalProviderMetadata, AuthType, ClientId, ClientSecret, IssuerUrl, ProviderMetadata,
+    Scope,
+};
+use std::env;
+
+use serde::{Deserialize, Serialize};
+
+use openidconnect::reqwest::http_client;
+
+use std::process::exit;
+
+// Obtain the device_authorization_url from the OIDC metadata provider.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+struct DeviceEndpointProviderMetadata {
+    device_authorization_endpoint: CoreDeviceAuthorizationUrl,
+}
+impl AdditionalProviderMetadata for DeviceEndpointProviderMetadata {}
+type DeviceProviderMetadata = ProviderMetadata<
+    DeviceEndpointProviderMetadata,
+    CoreAuthDisplay,
+    CoreClientAuthMethod,
+    CoreClaimName,
+    CoreClaimType,
+    CoreGrantType,
+    CoreJweContentEncryptionAlgorithm,
+    CoreJweKeyManagementAlgorithm,
+    CoreJwsSigningAlgorithm,
+    CoreJsonWebKeyType,
+    CoreJsonWebKeyUse,
+    CoreJsonWebKey,
+    CoreResponseMode,
+    CoreResponseType,
+    CoreSubjectIdentifierType,
+>;
+
+fn handle_error<T: std::error::Error>(fail: &T, msg: &'static str) {
+    let mut err_msg = format!("ERROR: {}", msg);
+    let mut cur_fail: Option<&dyn std::error::Error> = Some(fail);
+    while let Some(cause) = cur_fail {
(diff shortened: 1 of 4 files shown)


src/core/mod.rs
```diff
@@ -6,6 +6,10 @@ pub use oauth2::basic::{
     BasicRequestTokenError as CoreRequestTokenError,
     BasicRevocationErrorResponse as CoreRevocationErrorResponse, BasicTokenType as CoreTokenType,
 };
+
+use oauth2::devicecode::{DeviceAuthorizationResponse, EmptyExtraDeviceAuthorizationFields};
+pub use oauth2::DeviceAuthorizationUrl as CoreDeviceAuthorizationUrl;
```
ramosbugs, line 11: I think we can remove this since it's already being exported from the top-level `lib.rs` without the `Core` prefix and this isn't a trait or generic type that needs a typedef

src/lib.rs
```diff
@@ -1062,6 +1077,42 @@ where
...
+    ///
+    /// Creates a request builder for exchanging a device code for an access token.
+    /// This method is similar to [`exchange_device_code`][Client::exchange_device_code] but
+    /// allows the caller to specify additional parameters to be sent to the authorization server.
+    /// This is useful for providers that require additional parameters to be sent in the request
+    /// body.
+    ///
+    /// See https://tools.ietf.org/html/rfc8628#section-3.4
+    ///
+    pub fn exchange_device_token<'a, 'b, 'c>(
```
ramosbugs, line 1105: let's rename to `exchange_device_access_token` for consistency with `oauth2` and have it follow the same type signature (with generic `EF` so that users can access any vendor-specific fields that providers add). We'll need to re-export the `ExtraDeviceAuthorizationFields` trait above too (in addition to the `EmptyExtraDeviceAuthorizationFields` struct).

src/lib.rs
```diff
@@ -1062,6 +1077,42 @@ where
         self.oauth2_client.exchange_code(code)
     }
 
+    ///
+    /// Creates a request builder for exchanging a device code for an access token.
+    ///
+    /// See https://tools.ietf.org/html/rfc8628#section-3.4
```
ramosbugs, line 1083: let's linkify this URL by wrapping it in `<...>`. I think `cargo doc` warns when URLs don't have this. some of the other rustdocs in this crate have the same issue (not sure if rustdoc had that feature back then or if I was just unaware of it), but let's avoid new links that trigger warnings.

src/lib.rs
```diff
@@ -1062,6 +1077,42 @@ where
...
+    }
+
+    ///
+    /// Creates a request builder for exchanging a device code for an access token.
+    /// This method is similar to [`exchange_device_code`][Client::exchange_device_code] but
+    /// allows the caller to specify additional parameters to be sent to the authorization server.
+    /// This is useful for providers that require additional parameters to be sent in the request
+    /// body.
+    ///
+    /// See https://tools.ietf.org/html/rfc8628#section-3.4
```
ramosbugs, line 1103: linkify

src/lib.rs
```diff
@@ -1062,6 +1077,42 @@ where
...
+        if self.use_openid_scope {
+            Ok(request?.add_scope(Scope::new(OPENID_SCOPE.to_string())))
+        } else {
+            request
+        }
+    }
+
+    ///
+    /// Creates a request builder for exchanging a device code for an access token.
+    /// This method is similar to [`exchange_device_code`][Client::exchange_device_code] but
```
ramosbugs, line 1098: I'm confused about this comment. `exchange_device_access_code` and `exchange_device_token` are used sequentially for distinct purposes within the device flow; one isn't a more flexible version of the other, and both should allow setting any necessary parameters via their respective request builders.

src/lib.rs
```diff
@@ -576,7 +576,9 @@ extern crate pretty_assertions;
 #[macro_use]
 extern crate serde_derive;
 
+pub use oauth2::devicecode::{DeviceAuthorizationResponse, EmptyExtraDeviceAuthorizationFields};
```
ramosbugs, line 579: we should also re-export `DeviceCodeErrorResponseType`, `ExtraDeviceAuthorizationFields`, and `DeviceCodeErrorResponse` (see https://docs.rs/oauth2/4.4.0/oauth2/devicecode/index.html)

src/lib.rs
```diff
@@ -576,7 +576,9 @@ extern crate pretty_assertions;
...
 use oauth2::ResponseType as OAuth2ResponseType;
+pub use oauth2::{DeviceAccessTokenRequest, DeviceAuthorizationRequest};
```
ramosbugs, line 581: also `DeviceCode`

src/core/mod.rs
```diff
@@ -38,6 +42,12 @@ mod crypto;
 // Private purely for organizational reasons; exported publicly above.
 mod jwk;
 
+///
+/// OpenID Connect Core device authorization response.
```
ramosbugs, line 46: the OIDC spec doesn't deal with the device flow, so let's just copy the docs from oauth2:
> Standard implementation of DeviceAuthorizationResponse which throws away extra received response fields.

examples/okta_device_grant.rs
```diff
@@ -0,0 +1,135 @@
...
+            unreachable!();
+        });
+
+    // Use the custom metadata to get the device_authorization_endpoint
+    let device_authorization_endpoint = provider_metadata
+        .additional_metadata()
+        .device_authorization_endpoint
+        .clone();
+
+    // Set up the config for the Okta devuce authorization process.
```
ramosbugs, line 100: `device`

examples/okta_device_grant.rs
```diff
@@ -0,0 +1,135 @@
...
+    dbg!(&details);
+
+    // Display the URL and user-code.
+    println!(
+        "Open this URL in your browser:\n{}\nand enter the code: {}",
+        details
+            .verification_uri_complete()
+            .unwrap()
+            .secret()
+            .to_string(),
```
ramosbugs, line 121: nit: probably don't need to call `to_string()` here (or right below)

ramosbugs (commented):
thanks for the PR!

ramosbugs:
thanks!

PR #115, opened 2023-06-12
Implement `From<Url>` for `new_url_type`s

garryod:
I thought it would be nice to implement `From<Url>` for the url-like types. If there's a reason you believe this shouldn't be done (I note the existence of a `from_url` function) please feel free to close.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -332,6 +332,11 @@ macro_rules! new_url_type {
                 &self.1
             }
         }
+        impl From<Url> for $name {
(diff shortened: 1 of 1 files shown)


ramosbugs:
Hey @garryod, thanks for the PR. However, omitting `From<Url>` was an intentional design decision that follows the rationale I outlined for NewTypes in general here: https://github.com/ramosbugs/oauth2-rs/issues/112#issuecomment-675801545.

Issue #116, opened 2023-06-13
Dependencies update


ereOn:
Hi,

We are using `openidconnect` as part of our monorepo and also rely a lot on `cargo deny` to check/avoid/minimize duplicate dependencies.

I recently wanted to update the `openidconnect` crate from `2` to the latest `3` but noticed there are still a couple of direct dependencies that seem out-of-date. These end-up showing as duplicates and we have no other choice but to add exceptions for those. While not a deal breaker or anything, I still wonder whether there would be an opportunity to upgrade those instead of having to add exceptions.

I am talking about crates like `rsa`, `signature`, `serde_with` (to name a few) for which many newer releases have been available for some time (sometimes there is a 2 major versions difference) and which I suspect could be updated in `openidconnect`. 

Before I make a PR for this, is that something you guys would consider desirable ?

I understand - of course - that keeping always in-sync with the latest versions of each dependency is not a realistic goal, especially with a crate that does as much as `openidconnect`, but I feel that for many of those, updating would not be a huge risk or task. A low-hanging fruit if you will. I would add that using updated security crates (for `rsa`, `signature` and the like) could also be beneficial in terms of security.

What are your thoughts on this?

P.S: Thanks for all the good work on this crate: it is really great.

ramosbugs:
Hi @ereOn, I'd be happy to merge PRs that update dependencies without affecting any of this crate's public APIs. Any breaking changes should be discussed further, but I believe the crates you listed are only used internally.

ramosbugs:
@ereOn are there other dependencies still to update, or should we close this issue?

ereOn:
We can close this! Thanks again for considering this.

PR #117, opened 2023-06-19
Update dependencies to latest

ereOn:
Hi,

This PR updates some internally used dependencies to their latest version, in accordance to [this thread](https://github.com/ramosbugs/openidconnect-rs/issues/116).

I separated commits for the different crates updates, and each does compile individually. All tests pass at each intermediate step and obviously also at the final one.

Additionally, I took the liberty of fixing some minor annoyances:
- `cargo clippy` now runs without any warnings.
- `cargo test` now runs without any warnings (related to `chrono::Utc.timestamp()` being deprecated).

These are also part of different individual commits. Let me know if those last changes are not desired or desired as part of a different pull-request.

Thank you !


src/core/jwk.rs
```diff
@@ -449,82 +449,99 @@ impl
         signature_alg: &CoreJwsSigningAlgorithm,
         msg: &[u8],
     ) -> Result<Vec<u8>, SigningError> {
-        let (padding_alg, hash) = match *signature_alg {
+        match *signature_alg {
```
ereOn, line 452: The reason why I didn't keep the `match` & `return values` pattern here is because `padding_alg` is no longer an `enum` but a trait, and this would have forced us to `Box` it to return it here.

ramosbugs:
Thanks for the PR!

I think you'll need to update `Cargo-1.57.lock` for the MSRV CI step to pass:
```
cp Cargo-1.57.lock Cargo.lock
cargo check
cp Cargo.lock Cargo-1.57.lock
```

For context, the latest versions of some crates are incompatible with the current MSRV of Rust 1.57, so the repo has a lock file checked in to ensure that build step can pass consistently.

> Additionally, I took the liberty of fixing some minor annoyances:
> 
> * `cargo clippy` now runs without any warnings.
> * `cargo test` now runs without any warnings (related to `chrono::Utc.timestamp()` being deprecated).
> 
> These are also part of different individual commits. Let me know if those last changes are not desired or desired as part of a different pull-request.

Some of the clippy lints on the latest Rust release may suggest changes that use Rust features unavailable in 1.57. As long as the 1.57 build passes, feel free to address newer clippy lints, though. Since they're already in separate commits, I think it would be best to merge those in a separate PR from the version bumps.

ereOn:
Thanks for the feedback.

Just to clarify, is the compatibility with Rust 1.57 something that is still required ? This version is a year and a half old and I'm not quite sure to understand why in particular it needs to be supported.

AFAIK as I understand there is sadly no way for me to update to the latest versions of the crates and satisfy that constraint.

ramosbugs:
Which dependencies aren't compatible with 1.57? This crate's MSRV policy is to go back 6 months, which means it's ok to bump the MSRV to as recent as 1.66 (released December 2022). I would like to understand why it's needed, though. Compilation errors in dependencies seem like a valid reason to bump MSRV, but clippy lints and other compilation warnings do not.

ereOn:
The CI shows `p256` as not having candidate versions after `0.11.1` for rust `1.57`. This is confirmed by the fact that they indeed [bumped their Rust version criterion as of `0.12.0`](https://github.com/RustCrypto/elliptic-curves/blob/p256/v0.12.0/p256/Cargo.toml#L17).

I don't know yet if other crates are affected by similar changes.

ereOn:
I just realized I never tried out your suggestion with `cargo check`.

My apologies: I just did that and pushed the resulting commit.

Let's see what happens there.

ramosbugs:
It looks like 1.65 is the MSRV for these versions. I just added a commit to the PR to bump the MSRV for this crate accordingly.

ramosbugs:
Actually this requires some clippy lint updates, so I'll do that and push the MSRV update separately

ramosbugs:
I just pushed e9b1c53aa7f97cc9524ce5c8cb58c356b238ca07 to bump the MSRV. Rebasing should fix the compatibility issues, but you'll still need to update the 1.65 lock file to reflect the new dependency versions introduced in this PR (i.e., just rename your updated `Cargo-1.57.lock` to `Cargo-1.65.lock`)

ereOn:
@ramosbugs I just rebased my branch on top of `main` (which should contain the bump). I remove your commits from the branch as I understand they were already in `main` and the rebase included them anyway. Let me know if I got that wrong.

I redid the `cp/cargo check/cp` dance with the new lock file and pushed that as well.

Thanks for the quick reaction time!

ramosbugs:
Thanks!

ramosbugs:
This is now released in [3.3.0](https://crates.io/crates/openidconnect/3.3.0)

PR #121, opened 2023-07-16
Support Microsoft's style tenant id placeholders in issuer URLs

HannesSommer:
This adds all that is needed to work around MS `{tenantid}` placeholder in discovered issuer URLs.

Example: https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration

I'd be glad to learn about better solutions or to contribute unit test coverage if this change could be considered for integration.

### DIFF
--- a/src/discovery.rs
+++ b/src/discovery.rs
@@ -316,9 +316,14 @@ where
     ///
     /// Asynchronously fetches the OpenID Connect Discovery document and associated JSON Web Key Set
     /// from the OpenID Connect Provider.
+    /// It supports providing a custom expected issuer for IdPs that do not return exactly the same.
+    /// For instance, discovering Microsoft's OIDC configuration with the `common` tenant id at
+    /// https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration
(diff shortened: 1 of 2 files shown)


ramosbugs:
Unfortunately, Microsoft is egregiously violating the spec here with their use of `tenantid` placeholders in the discovery document, which matches neither the issuer URL nor the `iss` claim in the ID tokens. This was previously reported to them in https://github.com/MicrosoftDocs/azure-docs/issues/95994, but they don't seem to care (cc @nickludwig).

It's not reasonable to expect every OIDC library to special case Microsoft's non-compliant implementation, which defeats the whole purpose of having published standards in the first place. Unfortunately, Microsoft customers are on their own here and shouldn't expect out-of-the-box support from libraries that adhere to widely-used, published standards.

ramosbugs:
A suggested workaround is to provide a custom `http_client` function to `discover_async` that replaces the invalid` issuer` field in the discovery response with one that matches the issuer URL and `iss` claim subsequently returned in Microsoft's ID tokens, as [required](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata) in the spec.

PR #127, opened 2023-09-06
Add wasm support

Stygmates:
Chrono's wasm support wasn't enabled, probably because the [docs](https://docs.rs/chrono/0.4.29/chrono/index.html) were outdated and it is currently not listed in the crate's default features but is present in the [README](https://github.com/chronotope/chrono/blob/0.4.x/README.md) .

https://github.com/chronotope/chrono/pull/1260 updates the docs on their side. I can also make a PR for oauth2-rs if you want.

### DIFF
--- a/Cargo-1.65.lock
+++ b/Cargo-1.65.lock
@@ -149,9 +149,11 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "16b0a3d9ed01224b22057780a37bb8c5dbfe1be8ba48678e7bf57ec4b385411f"
 dependencies = [
  "iana-time-zone",
+ "js-sys",
  "num-integer",
  "num-traits",
  "serde",
+ "wasm-bindgen",
  "winapi",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ base64 = "0.13"
 chrono = { version = "0.4", default-features = false, features = [
     "clock",
     "std",
+    "wasmbind"
 ] }
 thiserror = "1.0"
 http = "0.2"


ramosbugs:
Interesting, what was the impact of not having this feature enabled? I would have assumed the WASM CI build step would have failed

Stygmates:
> Interesting, what was the impact of not having this feature enabled? I would have assumed the WASM CI build step would have failed

I ran into `panicked at 'time not implemented on this platform', library/std/src/sys/wasm/../unsupported/time.rs:31:9` that I think was raised when verifying the claims, specifically at this line: https://github.com/ramosbugs/openidconnect-rs/blob/414eb35419f84ab370c0d9acc3af72d72b67462c/src/verification.rs#L552.
I noticed that the implementation of `now()` used wasn't the correct one ([this one](https://github.com/chronotope/chrono/blob/cef90165dc00a3857130c4281fd68f0b69b3259c/src/offset/utc.rs#L90) which uses the std implementation of `now` which is not supported for wasm was used instead of [this one](https://github.com/chronotope/chrono/blob/cef90165dc00a3857130c4281fd68f0b69b3259c/src/offset/utc.rs#L105)) because of the feature flag

ramosbugs:
Oh ok, thanks! It looks this will require an update to the MSRV Cargo.lock file to get the 1.65 build to pass:
```
cp Cargo-1.65.lock Cargo.lock
cargo check
cp Cargo.lock Cargo-1.65.lock
```

Once that passes, I'll happily merge.

> I can also make a PR for oauth2-rs if you want.

That would be greatly appreciated 🙏

Stygmates:
This should pass now, I'll make the PR for oauth2-rs tomorrow, it's 1 AM where I live 😪

ramosbugs:
This is now released in [3.3.1](https://crates.io/crates/openidconnect/3.3.1)

Stygmates:
Thank you for the fast response and fast release of a new version!

PR #130, opened 2023-10-05
add EdDSA alg support

xshadowlegendx:
this pr contains

- [x] `eddsa` signing key option supports for `ed25519` variant
- [x] add `CoreJsonWebKey::new_okp`
- [x] add `CoreEdDsaPrivateSigningKey`
- [x] unit tests
  - [x] `test_eddsa_verification`
  - [x] `test_core_jwk_deserialization_ed`
  - [x]  `test_ed_signing` for `CoreEdDsaPrivateSigningKey`

discussion: [issue](https://github.com/ramosbugs/openidconnect-rs/issues/129)

lastly thanks everyone for all your awesome work on this library and happy to have any feedback about this pr

### DIFF
--- a/src/core/jwk.rs
+++ b/src/core/jwk.rs
@@ -165,6 +165,8 @@ impl CoreJsonWebKey {
             x: Some(Base64UrlEncodedBytes::new(x)),
             y: None,
             d: None,
+            #[cfg(feature = "jwk-alg")]
+            alg: None,
         }
     }
 }
@@ -541,6 +543,8 @@ impl
                 y: None,
                 d: None,
                 k: None,
+                #[cfg(feature = "jwk-alg")]
+                alg: None,
             },
         }
     }
@@ -1087,6 +1091,26 @@ mod tests {
             pkcs1_signing_input,
             signature_ed25519_other,
         );
+
+        // non-EdDsa key
+        if let Some(err) = key_ed25519
+            .verify_signature(
+                &CoreJwsSigningAlgorithm::EcdsaP256Sha256,
+                pkcs1_signing_input.as_bytes(),
+                signature_ed25519.as_bytes(),
+            )
+            .err()
+        {
+            let error_msg = "key type does not match signature algorithm".to_string();
+            match err {
+                SignatureVerificationError::InvalidKey(msg) => {
+                    if msg != error_msg {
+                        panic!("The error should be about key type")
+                    }
+                }
+                _ => panic!("We should fail before actual validation"),
+            }
+        }
     }
 
     #[test]
--- a/src/core/jwk.rs
+++ b/src/core/jwk.rs
@@ -437,7 +437,13 @@ impl
     }
 }
 
+///
+/// Variants of EdDSA Key.
+///
 pub enum EdDsaVariant {
+    ///
+    /// Ed25519 Variant
+    /// 
     Ed25519,
 }
 
--- a/src/core/jwk.rs
+++ b/src/core/jwk.rs
@@ -128,7 +128,7 @@ impl CoreJsonWebKey {
         }
     }
 
-    /// Instantiate a new ED public key from the raw x (`x`) part of the curve,
+    /// Instantiate a new Octet Key-Pair public key from the raw x (`x`) part of the curve,
     /// along with an optional (but recommended) key ID.
     ///
     /// The key ID is used for matching signed JSON Web Tokens with the keys used for verifying
--- a/src/core/jwk.rs
+++ b/src/core/jwk.rs
@@ -125,6 +125,30 @@ impl CoreJsonWebKey {
             d: None,
         }
     }
+
+    /// Instantiate a new ED public key from the raw x (`x`) part of the curve,
+    /// along with an optional (but recommended) key ID.
+    ///
+    /// The key ID is used for matching signed JSON Web Tokens with the keys used for verifying
+    /// their signatures.
+    pub fn new_okp(
+        x: Vec<u8>,
+        crv: CoreJsonCurveType,
+        kid: Option<JsonWebKeyId>
+    ) -> Self {
+        Self {
+            kty: CoreJsonWebKeyType::OctetKeyPair,
+            use_: Some(CoreJsonWebKeyUse::Signature),
+            kid,
+            n: None,
+            e: None,
+            k: None,
+            crv: Some(crv),
+            x: Some(Base64UrlEncodedBytes::new(x)),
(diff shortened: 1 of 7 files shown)


src/core/jwk.rs
```diff
@@ -402,6 +437,120 @@ impl
...
+///
+pub struct CoreEdDsaPrivateSigningKey {
+    kid: Option<JsonWebKeyId>,
+    key_pair: EdDsaSigningKey,
+}
+impl CoreEdDsaPrivateSigningKey {
+    ///
+    /// Converts an EdDSA private key (in PEM format) to a JWK representing its public key.
+    ///
+    pub fn from_pem(pem: &str, kid: Option<JsonWebKeyId>, variant: CoreEdDsaKeyVariant) -> Result<Self, String> {
```
ramosbugs, line 505: as above, `from_ed25519_pem`

src/core/jwk.rs
```diff
@@ -402,6 +437,120 @@ impl
...
+}
+impl
+    PrivateSigningKey<
+    CoreJwsSigningAlgorithm,
+    CoreJsonWebKeyType,
+    CoreJsonWebKeyUse,
+    CoreJsonWebKey,
+    > for CoreEdDsaPrivateSigningKey
+{
+    fn sign(&self, signature_alg: &CoreJwsSigningAlgorithm, message: &[u8]) -> Result<Vec<u8>, SigningError> {
```
ramosbugs, line 522: accepting a `signature_alg` here means we could (in the future) end up with a conflict between the key type (e.g., `EdDsaSigningKey::Ed25519`) and the `signature_alg` (e.g., Ed448). since, for EdDSA, the algorithm seems to be determined exclusively by the key type (vs. RSA that has different hash functions available to use with a given key), I think we should just remove the `signature_alg` argument. this also means the function will become infallible, and we no longer need to worry about an error type.

src/core/jwk.rs
```diff
@@ -402,6 +437,120 @@ impl
...
+    fn from_pem(pem: &str, variant: CoreEdDsaKeyVariant) -> Result<Self, String> {
+        match variant {
+            CoreEdDsaKeyVariant::Ed25519 => Ok(Self {
+                variant,
+                ed25519: Some(ed25519_dalek::SigningKey::from_pkcs8_pem(pem).map_err(|err| err.to_string())?),
+            })
+        }
+    }
+
+    fn public_part(&self) -> Vec<u8> {
```
ramosbugs, line 465: I think I would remove this function and just put the `match` in `as_verification_key()`, since that function needs to know both the curve type and the public key

ramosbugs (commented):
thanks for the PR!

let's also add `EdDSA (Ed25519 curve)` support to the README: https://github.com/ramosbugs/openidconnect-rs/blob/7b6fdd06ee4f728d7e4d40d256cccc5002178cb1/README.md#L37

I haven't reviewed the tests yet; will take a look at those after the current comments are resolved.

ramosbugs:
also, be sure to update the MSRV lock file `Cargo-1.65.lock` to address the failing 1.65.0 CI build:
```
cp Cargo-1.65.lock Cargo.lock
cargo check
cp Cargo.lock Cargo-1.65.lock
```

src/core/jwk.rs
xshadowlegendx: hello @ramosbugs, thanks for your reviews, you mean ignoring the argument right? since the `PrivateSigningKey` trait has the `signature_alg` parameter?

ramosbugs: oh! I missed that this was a trait impl 🤦‍♂️

nevermind, ignore my comment :-)

src/core/jwk.rs
```diff
@@ -863,6 +1019,41 @@ mod tests {
...
+            signature_ed448,
+        );
+
+        // different signature
+        verify_invalid_signature(
+            &key_ed25519,
+            &CoreJwsSigningAlgorithm::EdDsaEd25519,
+            pkcs1_signing_input,
+            signature_ed25519_other,
+        );
```
ramosbugs, line 1054: I think we probably also want to call `JsonWebKey::verify_signature()` with `CoreJwsSigningAlgorithm::EdDsaEd25519` on a non-EdDSA key, similar to: https://github.com/ramosbugs/openidconnect-rs/blob/67474002298c5b818adfcd9f74c898adf42758c6/src/core/jwk.rs#L968-L985

src/core/mod.rs
```diff
@@ -920,6 +927,11 @@ impl JwsSigningAlgorithm<CoreJsonWebKeyType> for CoreJwsSigningAlgorithm {
                 hasher.update(bytes);
                 hasher.finalize().to_vec()
             }
+            CoreJwsSigningAlgorithm::EdDsaEd25519 => {
```
ramosbugs, line 930: I believe it should be using SHA-512: https://www.rfc-editor.org/rfc/rfc8032#section-5.1. since that's the hash function used by the signature scheme, we should be using the same one here for computing `at_hash` and `c_hash`.

xshadowlegendx: hello @ramosbugs, I think the message is being hashed with `sha2-512` by default before signing with the key, I think it is the "prehash" from [section-5.1](https://www.rfc-editor.org/rfc/rfc8032#section-5.1) is implemented according to [this from ed25519_dalek](https://github.com/dalek-cryptography/ed25519-dalek/blob/main/src/signing.rs#L508)

ramosbugs: right, this code isn't used for signing JWTs. it's used for computing the [`at_hash`](https://openid.net/specs/openid-connect-core-1_0.html#CodeIDToken) and `c_hash` values, which is independent of the JWT signing and verification process.

ramosbugs: it just happens that the hash function used for computing those claims is supposed to be the same one that the signing algorithm uses, and in this case I believe that's SHA-512. otherwise, we wouldn't be able to compute these claims when using EdDSA, which doesn't seem right.

xshadowlegendx: oh is it we need to hash it again?

ramosbugs: Hopefully this helps clarify:
https://github.com/ramosbugs/openidconnect-rs/blob/67474002298c5b818adfcd9f74c898adf42758c6/src/id_token.rs#L87-L98

ramosbugs:
This is now released in [3.4.0](https://crates.io/crates/openidconnect/3.4.0)!

PR #131, opened 2023-10-06
Take JWK alg into account during key selection

multun:
Without this change, there may be multiple compatible keys, which causes an AmbiguousKeyId error. Using the key algorithm to narrow the list down further fixes the issue.

We've seen this edge case in the wild on a specific OpenAM instance.

### DIFF
--- a/src/core/jwk.rs
+++ b/src/core/jwk.rs
@@ -8,7 +8,7 @@ use crate::{
     PrivateSigningKey, SignatureVerificationError, SigningError,
 };
 
-use super::{crypto, CoreJwsSigningAlgorithm};
+use super::{crypto, CoreJwkAlgorithm, CoreJwsSigningAlgorithm};
 
 // Other than the 'kty' (key type) parameter, which must be present in all JWKs, Section 4 of RFC
 // 7517 states that "member names used for representing key parameters for different keys types
@@ -27,6 +27,11 @@ pub struct CoreJsonWebKey {
     pub(crate) use_: Option<CoreJsonWebKeyUse>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub(crate) kid: Option<JsonWebKeyId>,
+
+    /// The algorithm intended to be used with this key (https://www.rfc-editor.org/rfc/rfc7517#section-4.4)
+    /// It can either be an algorithm inteded for use with JWS or JWE, or something different.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub(crate) alg: Option<CoreJwkAlgorithm>,
 
     // From RFC 7517, Section 4: "Additional members can be present in the JWK; if not understood
     // by implementations encountering them, they MUST be ignored.  Member names used for
@@ -99,6 +104,7 @@ impl CoreJsonWebKey {
             x: None,
             y: None,
             d: None,
+            alg: None,
         }
     }
     /// Instantiate a new EC public key from the raw x (`x`) and y(`y`) part of the curve,
@@ -123,9 +129,11 @@ impl CoreJsonWebKey {
             x: Some(Base64UrlEncodedBytes::new(x)),
             y: Some(Base64UrlEncodedBytes::new(y)),
             d: None,
+            alg: None,
         }
     }
 }
+
 impl JsonWebKey<CoreJwsSigningAlgorithm, CoreJsonWebKeyType, CoreJsonWebKeyUse> for CoreJsonWebKey {
     fn key_id(&self) -> Option<&JsonWebKeyId> {
         self.kid.as_ref()
@@ -149,6 +157,7 @@ impl JsonWebKey<CoreJwsSigningAlgorithm, CoreJsonWebKeyType, CoreJsonWebKeyUse> 
             x: None,
             y: None,
             d: None,
+            alg: None,
         }
     }
 
@@ -159,19 +168,9 @@ impl JsonWebKey<CoreJwsSigningAlgorithm, CoreJsonWebKeyType, CoreJsonWebKeyUse> 
         signature: &[u8],
     ) -> Result<(), SignatureVerificationError> {
         use hmac::Mac;
-        if let Some(key_use) = self.key_use() {
-            if *key_use != CoreJsonWebKeyUse::Signature {
-                return Err(SignatureVerificationError::InvalidKey(
-                    "key usage not permitted for digital signatures".to_string(),
-                ));
-            }
-        }
 
-        if Some(self.key_type()) != signature_alg.key_type().as_ref() {
-            return Err(SignatureVerificationError::InvalidKey(
-                "key type does not match signature algorithm".to_string(),
-            ));
-        }
+        self.check_compatibility(signature_alg)
+            .map_err(|e| SignatureVerificationError::InvalidKey(e.to_owned()))?;
 
         match *signature_alg {
             CoreJwsSigningAlgorithm::RsaSsaPkcs1V15Sha256 => {
@@ -323,6 +322,35 @@ impl JsonWebKey<CoreJwsSigningAlgorithm, CoreJsonWebKeyType, CoreJsonWebKeyUse> 
                     )
                 }),
             )),
+        }
+    }
+
+    fn check_compatibility(
+        &self,
+        signing_algorithm: &CoreJwsSigningAlgorithm,
+    ) -> Result<(), &'static str> {
+        // if this key isn't suitable for signing
+        if let Some(use_) = &self.use_ {
+            if !use_.allows_signature() {
+                return Err("key usage not permitted for digital signatures");
+            }
+        }
+
+        // if this key doesn't have the right key type
+        if signing_algorithm.key_type().as_ref() != Some(self.key_type()) {
+            return Err("key type does not match signature algorithm");
+        }
+
+        // if no specific algorithm is mandated, any will do
+        let key_alg = match self.alg.clone() {
+            None => return Ok(()),
+            Some(key_alg) => key_alg,
+        };
+
+        match key_alg {
+            CoreJwkAlgorithm::JWS(ref key_alg) if key_alg == signing_algorithm => Ok(()),
+            CoreJwkAlgorithm::JWS(_) => Err("incompatible key algorithm"),
+            _ => Err("key algorithm is not a signing algorithm"),
         }
     }
 }
@@ -554,6 +582,7 @@ impl
             x: None,
             y: None,
             d: None,
+            alg: None,
         }
     }
 }
@@ -1121,21 +1150,6 @@ mod tests {
             signing_input,
             "s0h6KThzkfBBBkLspW1h84VsJZFTsPPqMDA7g1Md7p0",
         );
-
-        verify_signature(
-            &key,
-            &CoreJwsSigningAlgorithm::HmacSha384,
-            signing_input,
-            "O1jhTTHkuaiubwDZoIBLv6zjEarXHc22NNu05IdYh_yzIKGYXJQcaI2WnF4BCq7j",
-        );
-
-        verify_signature(
-            &key,
-            &CoreJwsSigningAlgorithm::HmacSha512,
-            signing_input,
-            "rdWYqzXuAJp4OW-exqIwrO8HJJQDYu0_fkTIUBHmyHMFJ0pVe7fjP7QtE7BaX-7FN5\
-             YiyiM11MwIEAxzxBj6qw",
-        );
     }
 
     fn expect_hmac(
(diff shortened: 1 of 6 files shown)


ramosbugs:
Hey, thanks for the PR. According to the [spec](https://openid.net/specs/openid-connect-core-1_0.html#Signing), this issue should never arise:
> If there are multiple keys in the referenced JWK Set document, a `kid` value MUST be provided in the JOSE Header.

I'm hesitant to remove this `MUST` requirement in case there are clever ways for an attacker to take advantage of key confusion. This is similar to past high-profile JWT vulnerabilities that allowed attackers to change the `alg` value in a JWT to `HS256` to coerce a library into treating an RSA public key as as a symmetric secret. I don't *think* this particular change will introduce security issues, but the reason standards exist is so that individual developers don't have to guess about these things without review from other experts.

Is this issue popping up with a popular OIDC provider?

multun:
Hey! thanks for the review

> I'm hesitant to remove this MUST requirement in case there are clever ways for an attacker to take advantage of key confusion. 

No requirement is removed, as a kid is indeed always provided: it's just not unique, there are multiple keys with the same kid but a different alg field. `kid`s [SHOULD be unique](https://www.rfc-editor.org/rfc/rfc7517.html#section-4.5) but are explicitly allowed not to.

The specification gives an example for key types, which is correctly handled by the current code: you can have an RSA and an EC key with the same `kid`, and `openidconnect-rs` will pick the right one just fine. Another such case which is already handled by the current code is key usage: you could have two keys with the same `kid`, one for signing, and one for encryption. I'd argue that honoring the `alg` field isn't different.

This is an [optional setting](https://backstage.forgerock.com/docs/am/7.2/oidc1-guide/managing-jwk_uri.html#kid-multiple-keys) of OpenAM, called `Include all kty and alg combinations in jwk_uri` and it's there for a good reason: It's designed to enforce a list of allowed algorithms per key, which enables deprecating vulnerable algorithms:

```json
[
    {
        "kty": "RSA",
        "kid": "eGxIf8PmOKJeqN/jPAj85u6LMB0=",
        "use": "sig",
        "n": "rV7qnTenVeZwYP4m0ji5xbsFvqNRTW8-i0Ys98oBVKlWJpEXdPUQMBi4TdjHPUBRmueW74v1v8Uw-1NeE8WvI0bStvH7P2zxeP5bdL6onVNIZUdb1L1lkBjlYQP30TtsRZuJ2d-vmf3BKEvtd3V47A8gSuAJO9q8dT-Rby7ZOMWw_ZU_dJTIGplhgpJlQMXi3wLZyHU-oi7V5PmRE0ZYEn0LLXtQXQj1bYW-5AjU6TykXQVqISqImGiONpnKQYkOgZ56vXR9nU-_ZSmyc_VTBTnA0Xwj_aWfOokaFqft0LhH1gykhq9IIgHaxo55SqRm4lymxx13Hpe1lA3BlWzWVQ",
        "e": "AQAB",
        "alg": ["RS512", "PS256"]
    }
]
```

Of course, the example above does not work, as alg isn't a set of allowed algorithms, so it has to be de-normalized:

```json
[
    {
        "kty": "RSA",
        "kid": "eGxIf8PmOKJeqN/jPAj85u6LMB0=",
        "use": "sig",
        "n": "rV7qnTenVeZwYP4m0ji5xbsFvqNRTW8-i0Ys98oBVKlWJpEXdPUQMBi4TdjHPUBRmueW74v1v8Uw-1NeE8WvI0bStvH7P2zxeP5bdL6onVNIZUdb1L1lkBjlYQP30TtsRZuJ2d-vmf3BKEvtd3V47A8gSuAJO9q8dT-Rby7ZOMWw_ZU_dJTIGplhgpJlQMXi3wLZyHU-oi7V5PmRE0ZYEn0LLXtQXQj1bYW-5AjU6TykXQVqISqImGiONpnKQYkOgZ56vXR9nU-_ZSmyc_VTBTnA0Xwj_aWfOokaFqft0LhH1gykhq9IIgHaxo55SqRm4lymxx13Hpe1lA3BlWzWVQ",
        "e": "AQAB",
        "alg": "RS512"
    },
    {
        "kty": "RSA",
        "kid": "eGxIf8PmOKJeqN/jPAj85u6LMB0=",
        "use": "sig",
        "n": "rV7qnTenVeZwYP4m0ji5xbsFvqNRTW8-i0Ys98oBVKlWJpEXdPUQMBi4TdjHPUBRmueW74v1v8Uw-1NeE8WvI0bStvH7P2zxeP5bdL6onVNIZUdb1L1lkBjlYQP30TtsRZuJ2d-vmf3BKEvtd3V47A8gSuAJO9q8dT-Rby7ZOMWw_ZU_dJTIGplhgpJlQMXi3wLZyHU-oi7V5PmRE0ZYEn0LLXtQXQj1bYW-5AjU6TykXQVqISqImGiONpnKQYkOgZ56vXR9nU-_ZSmyc_VTBTnA0Xwj_aWfOokaFqft0LhH1gykhq9IIgHaxo55SqRm4lymxx13Hpe1lA3BlWzWVQ",
        "e": "AQAB",
        "alg": "PS256"
    }
]
```
  
> This is similar to past high-profile JWT vulnerabilities that allowed attackers to change the alg value in a JWT to HS256 to coerce a library into treating an RSA public key as as a symmetric secret. I don't think this particular change will introduce security issues, but the reason standards exist is so that individual developers don't have to guess about these things without review from other experts.

I doubt this change could introduce key confusion issues: it makes key selection stricter by honoring the key's `alg` field.

> Is this issue popping up with a popular OIDC provider?

It depends on what you mean by provider:
 - The software used on the server side is OpenAM, which I believe counts as popular
 - This OP instance is not even public, it's some internal stuff for my company

src/core/mod.rs
```diff
@@ -933,6 +933,29 @@ impl JwsSigningAlgorithm<CoreJsonWebKeyType> for CoreJwsSigningAlgorithm {
...
+/// There are a few caveats, which can be deduced from [the "Algorithm Usage Location(s)" part of Section 7.1.1](https://www.rfc-editor.org/rfc/rfc7518#section-7.1.1):
+///  - According to the specification, algorithms might have multiple purposes (like both JWE and JWS).
+///    No such algorithm is currently listed, but it if were to happen, the current flattened enum approach would stop working
+///  - According to the specification, algorithms which aren't used for JWE nor JWS may be added to the list.
+///    No such algorithm is currently listed.
+/// According to the "Algorithm Usage Location(s)" section of  https://www.rfc-editor.org/rfc/rfc7518#section-7.1.1
+///
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(untagged)]
+pub enum CoreJwkAlgorithm {
```
ramosbugs, line 950: this type serves a similar purpose as `crate::jwt::JsonWebTokenAlgorithm`, with the addition of the key management algorithms. it doesn't seem like the key management algorithms will ever actually be included in a JWK, at least within the context of OIDC, since these are typically private keys already held by the client and used to decrypt the CEK (iiuc, please correct me if I'm wrong). I think it should be fine to use `JsonWebTokenAlgorithm` instead of introducing a new type. the `deserialize_option_or_none` suggestion above will also prevent deserialization errors in the event that algorithms not understood by this crate (including JWE key management algorithms) are used.

as far as handling algorithms that could be used by both JWE and JWS, the "proper" solution is probably to look at the `use` field (if provided) to disambiguate. but, that would require a more complex `Deserialize` implementation that isn't worth adding at this point.

src/types.rs
```diff
@@ -233,6 +233,12 @@ where
     ///
     fn key_use(&self) -> Option<&JU>;
 
+    ///
+    /// Returns the algorithm (e.g. ES512) this key must be used with, or `None` if no algorithm
+    /// constraint was specified.
+    ///
+    fn check_compatibility(&self, signing_algorithm: &JS) -> Result<(), &'static str>;
```
ramosbugs, line 240: this is a breaking change to a public trait, but I don't see a way to implement the desired functionality without adding a new method to the trait. to avoid a major version bump, let's put this behind a new optional feature flag called `jwk-alg` or something.

for simplicity and consistency with the rest of this trait, I think I'd prefer this function to be:
```rust
fn signing_alg(&self) -> Option<&JS>;
```
with that API, I think the rest of the logic can be folded into `filter_keys`

src/types.rs
```diff
@@ -754,6 +760,24 @@ where
         }
     }
 
+    ///
+    /// Return a list of suitable keys, given a key id an signature algorithm
+    ///
+    pub fn filter_keys(&self, key_id: &Option<JsonWebKeyId>, signature_alg: &JS) -> Vec<&K> {
```
ramosbugs, line 766: I'd prefer not to add this to the stable public API. let's just make it `pub(crate)` for now.

ramosbugs (commented):
thanks for the additional explanation! this change makes sense now and seems consistent with the standards.

src/core/jwk.rs
```diff
@@ -28,6 +28,11 @@ pub struct CoreJsonWebKey {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub(crate) kid: Option<JsonWebKeyId>,
 
+    /// The algorithm intended to be used with this key (https://www.rfc-editor.org/rfc/rfc7517#section-4.4)
+    /// It can either be an algorithm inteded for use with JWS or JWE, or something different.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
```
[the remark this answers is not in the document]
ramosbugs: on second thought, let's put this behind the feature flag as well (without `deserialize_option_or_none`), and the existing `VecSkipError`  usage in `JsonWebKeySet` will skip any keys with unrecognized algorithms (when the feature is enabled), which we shouldn't use anyway if this crate doesn't understand them: https://github.com/ramosbugs/openidconnect-rs/blob/1bc4a2f6c741f1c94716c5f4d897b0f3eaa06fbe/src/types.rs#L735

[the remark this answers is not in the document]
multun: yup, we're on the same page

> should make it a non-breaking change that will ignore unrecognized key types

That's what I did at first, but then we couldn't differentiate keys with an unknown alg from keys with no alg specified. I also tried adding an Unknown member to the enum (marked with `serde(other)`), but it made the change very intrusive. So I settled on relying on VecSkipError to skip keys with unknown algs at parsing.

>  on second thought, let's put this behind the feature flag as well (without deserialize_option_or_none), and the existing VecSkipError usage in JsonWebKeySet will skip any keys with unrecognized algorithms

Sounds good! Do you have a preferred name for the feature flag? I understand that you don't want to make a major release just for that, I'm guessing the feature flag would be removed on the next major release?

src/core/mod.rs
multun: > it doesn't seem like the key management algorithms will ever actually be included in a JWK, at least within the context of OIDC, since these are typically private keys already held by the client and used to decrypt the CEK (iiuc, please correct me if I'm wrong)

I had no idea which ones made sense to include, so defaulted back to doing as the spec said, which wasn't as clear as I wish it were. 

> I think it should be fine to use JsonWebTokenAlgorithm instead of introducing a new type.

I'm happy to use the existing type, I looked for it but failed to find it

src/types.rs
multun: > for simplicity and consistency with the rest of this trait, I think I'd prefer this function to be: fn signing_alg(&self) -> Option<&JS>;

That's what I did at first, but I believe there's an issue: with this API, we can't differentiate a missing alg field from an alg field with an encryption algorithm. I think it's an issue, as if the key does not have a key use, but has an encryption `alg`, we probably shouldn't use it for signing. I can think of multiple ways to address this issue:
 - return an `Option<Option<&JS>>`, where `None` is a missing field and `Some(None)` is an encryption algorithm
 - return an `Option<&KeyAlgorithm<JS>>`, with `enum KeyAlgorithm<JS> { None, Signature(JS), Encryption }`
 - return an `Option<&KeyAlgorithm<JS, JE>>`, with `enum KeyAlgorithm<JS, JE> { None, Signature(JS), Encryption(JE) }`. this is my favorite one, but it requires adding one more generic parameter to the JsonWebKey trait, which would require a major version bump
 - let the trait implementer decide (which is what I did)

ramosbugs: ah yes, that makes sense (`None` being ambiguous). what if we define a simpler enum for this function to return:
```rust
pub enum JsonWebKeySigningAlgorithm<JS, JT>
where
    JS: JwsSigningAlgorithm<JT>,
    JT: JsonWebKeyType,
{
  // sigh, PhantomData seems unavoidable here, as it would be in KeyAlgorithm
  Algorithm(JS, PhantomData<JT>),
  Unspecified,
  Unsupported,
}
```
the `Unspecified` variant instructs this library to allow the key to be used with any compatible signature algorithm (assuming the key use also allows signatures), while the `Unsupported` variant means the associated algorithm doesn't support signing (i.e., because it's a JWE algorithm). this will accomplish a similar purpose as `KeyAlgorithm` but without introducing a new type parameter. I agree that if we didn't have to worry about backward compatibility, `KeyAlgorithm` feels like the most natural approach (and may be worth adopting in a future major version).

src/core/jwk.rs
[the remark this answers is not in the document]
ramosbugs: >  I understand that you don't want to make a major release just for that, I'm guessing the feature flag would be removed on the next major release?

yup, hopefully I'll remember to remove it in the next major release. it's probably worth adding a TODO comment next to the feature flag's definition in `Cargo.toml`.

also we can remove the `default` within `serde()` since it's implied by `Option<_>` in the absence of a custom deserializer.

src/core/mod.rs
ramosbugs: > I'll trust you on this one: You have a lot more expertise than me on the matter, I had no idea whether key management algorithms belonged there.

I think technically any key can be formatted as a JWK, but I wouldn't expect to find a key management one in a JWK set returned by the provider's `jwks_uri`, which is the primary (sole?) purpose of the `JsonWebKey` type in this crate. in any case, this crate probably wouldn't have use for such a key (so it's fine to ignore via `VecSkipError`).

src/types.rs
multun: Yep, that was option 2 on my list, I'm happy to make that change

ramosbugs: sounds good!

ramosbugs: > that was option 2 on my list

I don't think it should return an option though since I'm not sure what `None` would mean in that context

multun: indeed, my bad

multun:
@ramosbugs I've done my best to take your review into account. A few things worth mentioning:
 - [One of the tests used a key management algorithm in a key](https://github.com/ramosbugs/openidconnect-rs/pull/131/files#diff-e8b6a8989304a97661be9a4861793b15823b2bb4eec2ef13024511695c1fbd72R789)
 - I retained the check_key_compatibility function, as it's used both when [filtering keys](https://github.com/ramosbugs/openidconnect-rs/pull/131/files#diff-ed12f5ea605f23bb4d26cc65778e3daff9db3eec40e2abfc82beafca130a99a0R822) and when [verifying signatures](https://github.com/ramosbugs/openidconnect-rs/pull/131/files#diff-e8b6a8989304a97661be9a4861793b15823b2bb4eec2ef13024511695c1fbd72R186)

src/core/jwk.rs
```diff
@@ -1477,4 +1497,83 @@ mod tests {
...
+            ]
+        }";
+        let jwks = serde_json::from_str::<CoreJsonWebKeySet>(jwks_json)
+            .expect("deserialization should succeed");
+
+        let keys = jwks.filter_keys(
+            &Some(JsonWebKeyId::new("2011-04-29".to_string())),
+            &CoreJwsSigningAlgorithm::RsaSsaPssSha384,
+        );
+        assert_eq!(keys.len(), 1);
```
ramosbugs, line 1570: I think we should also assert that the JWKS started with 2 keys in case one got filtered out by `VecSkipError` instead of `filter_keys()`. and then maybe also call `filter_keys()` with `PS512` and assert that it returns an empty set.

src/core/mod.rs
```diff
@@ -933,6 +933,7 @@ impl JwsSigningAlgorithm<CoreJsonWebKeyType> for CoreJwsSigningAlgorithm {
     }
 }
 
+
```
ramosbugs, line 936: I think `cargo fmt` failed here in the 1.65.0 build

src/core/jwk.rs
```diff
@@ -1121,21 +1156,6 @@ mod tests {
             signing_input,
             "s0h6KThzkfBBBkLspW1h84VsJZFTsPPqMDA7g1Md7p0",
         );
-
```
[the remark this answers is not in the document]
multun: The key has [the alg field set to `HS256`](https://github.com/ramosbugs/openidconnect-rs/pull/131/files#diff-e8b6a8989304a97661be9a4861793b15823b2bb4eec2ef13024511695c1fbd72R1140), so we can't verify the signature with anything else

ramosbugs (approved):

ramosbugs:
Thanks! I'll probably cut a release in the next week or so, once I have a chance to review and merge #130.

ramosbugs:
This is now released in [3.4.0](https://crates.io/crates/openidconnect/3.4.0)!

PR #142, opened 2023-12-26
Include the full discovery url on invalid HTTP response

Timshel:
Hey,

Simple PR to add the full discovery url to the error since in lot of cases it will be due to an improper `issuer_url`.

### DIFF
--- a/src/discovery.rs
+++ b/src/discovery.rs
@@ -302,9 +302,11 @@ where
             .join(CONFIG_URL_SUFFIX)
             .map_err(DiscoveryError::UrlParse)?;
 
-        http_client(Self::discovery_request(discovery_url))
+        http_client(Self::discovery_request(discovery_url.clone()))
             .map_err(DiscoveryError::Request)
-            .and_then(|http_response| Self::discovery_response(issuer_url, http_response))
+            .and_then(|http_response| {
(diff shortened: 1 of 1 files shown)


ramosbugs:
Thanks for the PR! I'll merge once CI passes (looks like `cargo fmt` is unhappy)

Timshel:
O yes sorry :)

Timshel:
I had tried running it and it failed so I invalidly concluded that you did not use it.
On another project it's included in the `rust-toolchain.toml`.

But I'm a bit of a rust noob so not sure of the best practices :).

ramosbugs:
Thanks!

> I had tried running it and it failed so I invalidly concluded that you did not use it.

When I run it locally with Cargo/Rust 1.71.0 (`rustfmt 1.5.2-stable`), it doesn't seem to change any of the formatting. Is it possible you ran it with the nightly toolchain?

> On another project it's included in the `rust-toolchain.toml`.

I think this file is mainly for pinning to specific versions: https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file.

ramosbugs:
This is now released in [3.5.0](https://crates.io/crates/openidconnect/3.5.0).

PR #146, opened 2024-01-30
feat: Add `UserInfoRequest::set_response_type`

### LINKED ISSUE #145: change UserInfo request's Accept header
Hi, thanks for this awesome crate! I'm not sure if this is just another instance of the server we are interacting not being spec compliant again, but when we're sending the request to the UserInfo endpoint (https://docs.rs/openidconnect/3.4.0/openidconnect/struct.Client.html#method.user_info) the server will return a 302 to a 404 (this crate stops at the 302) because it checks the Accept header set [here](https://github.com/ramosbugs/openidconnect-rs/blob/502ad930da6aa8451cfc262be99be681a6a1730c/src/user_info.rs#L101).

I need to either remove the header, or set it to `MIME_TYPE_JWT`. I confirmed in a local copy of this crate that this will make it work for us. As far as i understand it, it also makes sense to be able to set JWT since the crate can handle JWT responses.

Would this be in scope for this crate? Or is this simply the server's fault and we have to deal with that?

Thanks in advance :)

FabianLars:
closes #145 

Does this come close to what you had in mind? I'm not good at writing docs so apologies for the wording in advance.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -647,7 +647,8 @@ pub use types::{
 };
 
 pub use user_info::{
-    UserInfoClaims, UserInfoError, UserInfoJsonWebToken, UserInfoRequest, UserInfoUrl,
+    UserInfoClaims, UserInfoError, UserInfoJsonWebToken, UserInfoRequest, UserInfoResponseType,
+    UserInfoUrl,
 };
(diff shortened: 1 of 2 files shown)


ramosbugs:
this is great, thank you!

PR #158, opened 2024-04-14
Return `impl Future` instead of `Pin<Box<dyn Future>>`

sbihel:
Related to [ramosbugs/oauth2#6e583bd](https://github.com/ramosbugs/oauth2-rs/commit/6e583bd03203e42ef712fc90edb57cf5a389f9b7)


ramosbugs:
Thank you! 🙏

PR #161, opened 2024-04-17
Add support for specific JOSE header types

### LINKED ISSUE #160: `typ` validation in JWT JOSE header
Currently, in the [verification module](https://github.com/ramosbugs/openidconnect-rs/blob/c4e28f43bdd5faba098e92400476670ff057bf8e/src/verification/mod.rs#L205) the `typ` header field of the JOSE header of JWT is checked.

However, I am not sure why exactly as I did not find any spec warranting [the comment in line 205](https://github.com/ramosbugs/openidconnect-rs/blob/c4e28f43bdd5faba098e92400476670ff057bf8e/src/verification/mod.rs#L205).

I checked [RFC 7515](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.9) stating "Use of this Header Parameter is OPTIONAL" and even more that the `typ` can even be `JOSE` etc. Also [RFC 7519](https://www.rfc-editor.org/rfc/rfc7519#section-5.1) mentions that even if the header is set it is just recommended that the header should be set to "JWT".

At the moment, there is no possibility builtin to turn off or change the verification. Personally, I would see three different options to improve the situation:

1. Remove this check altogether. It has no security implications and the checks are only recommended. (It is already skipped if the claim is not set at all)
2. Allow disabling the check
3. Allow specifying a list of allowed `typ` values while defaulting to `JWT`

lmm-git:
Check https://www.rfc-editor.org/rfc/rfc7519#section-5.1 and https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.9 for more information on when the typ header should be set and to which value.

This commit allows to skip the check altogether as warranted by RFC 7515 the use of the header is optional but if set, according to RFC 7519, it should be set to "JWT". For other token types this is different (see RFC 9068).

Resolves #160

### DIFF
--- a/src/jwt/mod.rs
+++ b/src/jwt/mod.rs
@@ -21,6 +21,7 @@ new_type![
 ];
 
 new_type![
+    /// JSON web token type field (typ)
     #[derive(Deserialize, Hash, Ord, PartialOrd, Serialize)]
     JsonWebTokenType(String)
 ];
--- a/src/jwt/mod.rs
+++ b/src/jwt/mod.rs
@@ -20,11 +20,107 @@ new_type![
     JsonWebTokenContentType(String)
 ];
 
+/// Error type used when normalizing [`JsonWebTokenType`] objects
+#[derive(Error, Debug)]
+#[error("Invalid JWT type: {typ}")]
+pub struct InvalidJsonWebTokenTypeError {
+    typ: String,
+}
+
 new_type![
-    /// JSON web token type field (typ)
-    #[derive(Deserialize, Hash, Ord, PartialOrd, Serialize)]
+    /// JSON Web Token type field (typ)
+    ///
+    /// This type stores the raw (deserialized) value.
+    ///
+    /// To compare two different JSON Web Token types, please use the normalized version via [`JsonWebTokenType::normalize`].
+    #[derive(Deserialize, Hash, Serialize)]
     JsonWebTokenType(String)
+
+    impl {
+        /// Expands a [`JsonWebTokenType`] and produces a [`NormalizedJsonWebTokenType`] according to RFC2045 and RFC7515.
+        ///
+        /// See [RFC 2045 section 5.1](https://tools.ietf.org/html/rfc2045#section-5.1) for the full Content-Type Header Field spec.
+        /// See [RFC 7515 section 4.19](https://tools.ietf.org/html/rfc7515#section-4.1.9) for specific requirements of JSON Web Token Types.
+        pub fn normalize(&self) -> Result<NormalizedJsonWebTokenType, InvalidJsonWebTokenTypeError> {
+            self.try_into()
+        }
+    }
 ];
+
+/// Normalized JSON Web Token type field (typ)
+///
+/// This type stores the normalized value of a [`JsonWebTokenType`].
+/// To retrieve a normalized value according to RFC2045 and RFC7515 see [`JsonWebTokenType::normalize`]
+///
+/// See [RFC 2045 section 5.1](https://tools.ietf.org/html/rfc2045#section-5.1) for the full Content-Type Header Field spec.
+/// See [RFC 7515 section 4.1.9](https://tools.ietf.org/html/rfc7515#section-4.1.9) for specific requirements of JSON Web Token Types.
+///
+/// It is recommended to instantiate `NormalizedJsonWebTokenType` objects via [`JsonWebTokenType`] and then call [`JsonWebTokenType::normalize`].
+///
+/// ```rust
+/// # use openidconnect::{NormalizedJsonWebTokenType, JsonWebTokenType};
+/// let token_type = JsonWebTokenType::new("jwt+at".to_string()).normalize();
+/// // normalized value looks like "application/jwt+at"
+/// # assert_eq!(*token_type.unwrap(), "application/jwt+at")
+/// ```
+#[derive(Clone, Debug, Deserialize, Hash, Eq, PartialEq, Ord, PartialOrd, Serialize)]
+pub struct NormalizedJsonWebTokenType(String);
+
+impl std::ops::Deref for NormalizedJsonWebTokenType {
+    type Target = String;
+    fn deref(&self) -> &String {
+        &self.0
+    }
+}
+impl From<NormalizedJsonWebTokenType> for String {
+    fn from(t: NormalizedJsonWebTokenType) -> String {
+        t.0
+    }
+}
+
+impl From<NormalizedJsonWebTokenType> for JsonWebTokenType {
+    fn from(t: NormalizedJsonWebTokenType) -> JsonWebTokenType {
+        JsonWebTokenType::new(t.0)
+    }
+}
+
+impl TryFrom<&JsonWebTokenType> for NormalizedJsonWebTokenType {
+    type Error = InvalidJsonWebTokenTypeError;
+
+    /// Normalizes a [`JsonWebTokenType`] and produces a [`NormalizedJsonWebTokenType`] according to RFC2045.
+    ///
+    /// See [RFC 2045 section 5.1](https://tools.ietf.org/html/rfc2045#section-5.1) for the full Content-Type Header Field spec.
+    /// See [RFC 7515 section 4.19](https://tools.ietf.org/html/rfc7515#section-4.1.9) for specific requirements of JSON Web Token Types.
+    fn try_from(t: &JsonWebTokenType) -> Result<NormalizedJsonWebTokenType, Self::Error> {
+        let lowercase_jwt_type = t.0.to_lowercase();
+        if let Some(slash_location) = lowercase_jwt_type.find('/') {
+            if let Some(semicolon_location) = lowercase_jwt_type.find(';') {
+                // If '/' is not before ';' as then the MIME type is invalid
+                // e.g. some;arg="1/2" is invalid, but application/some;arg=1 is valid
+                // OR
+                // If MIME type has not at least one character
+                // OR
+                // If MIME subtype has not at least one character
+                if slash_location > semicolon_location
+                    || slash_location == 0
+                    || slash_location.saturating_add(1) >= semicolon_location
+                {
+                    Err(InvalidJsonWebTokenTypeError {
+                        typ: lowercase_jwt_type,
+                    })
+                } else {
+                    Ok(NormalizedJsonWebTokenType(lowercase_jwt_type))
+                }
+            } else {
+                Ok(NormalizedJsonWebTokenType(lowercase_jwt_type))
+            }
+        } else {
+            Ok(NormalizedJsonWebTokenType(format!(
+                "application/{lowercase_jwt_type}"
+            )))
+        }
+    }
+}
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 #[non_exhaustive]
--- a/src/verification/mod.rs
+++ b/src/verification/mod.rs
@@ -1,10 +1,10 @@
-use crate::jwt::{JsonWebToken, JsonWebTokenJsonPayloadSerde, JsonWebTokenType};
+use crate::jwt::{JsonWebToken, JsonWebTokenJsonPayloadSerde, NormalizedJsonWebTokenType};
 use crate::user_info::UserInfoClaimsImpl;
 use crate::{
     AdditionalClaims, Audience, AuthenticationContextClass, ClientId, ClientSecret, GenderClaim,
     IdTokenClaims, IssuerUrl, JsonWebKey, JsonWebKeyId, JsonWebKeySet, JsonWebTokenAccess,
-    JsonWebTokenAlgorithm, JsonWebTokenHeader, JweContentEncryptionAlgorithm, JwsSigningAlgorithm,
-    Nonce, SubjectIdentifier,
+    JsonWebTokenAlgorithm, JsonWebTokenHeader, JsonWebTokenType, JweContentEncryptionAlgorithm,
+    JwsSigningAlgorithm, Nonce, SubjectIdentifier,
 };
 
 use chrono::{DateTime, Utc};
@@ -118,7 +118,7 @@ pub(crate) struct JwtClaimsVerifier<'a, K>
 where
     K: JsonWebKey,
 {
-    allowed_jose_types: Option<HashSet<JsonWebTokenType>>,
+    allowed_jose_types: Option<HashSet<NormalizedJsonWebTokenType>>,
     allowed_algs: Option<HashSet<K::SigningAlgorithm>>,
     aud_match_required: bool,
     client_id: ClientId,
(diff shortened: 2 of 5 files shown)


src/verification/mod.rs
```diff
@@ -118,6 +118,7 @@ pub(crate) struct JwtClaimsVerifier<'a, K>
 where
     K: JsonWebKey,
 {
+    allowed_jose_types: Option<Vec<JsonWebTokenType>>,
```
ramosbugs, line 121: based on my comment in #160, let's just change this to `is_typ_check_enabled: bool`

ramosbugs (commented):
thanks for the PR, and for adding tests!

ramosbugs:
Superceded by #175

lmm-git:
Hm, this is a bit disappointing as I waited for your review for three weeks :(

Also the other MR does not cover the needs for verifying other types correctly.

ramosbugs:
Hey @lmm-git, I saw my unresolved comment about `is_typ_check_enabled` above but missed that I had agreed to merge option 3 in #160. Apologies for the miscommunication and delay (have a newborn at home as of ~3 weeks ago). I'll reopen this PR and the associated issue.

lmm-git:
Thanks for your response!

How do we want to proceed now? My proposal would be to revert the commit from #175 on this branch and apply the already existing commit. Or do you like it more if I rebase on top of the current main?

src/verification/mod.rs
ramosbugs, line 121: since these should be unique, let's use a `HashSet` (following the example of `allowed_algs`)

src/verification/mod.rs
```diff
@@ -181,6 +187,27 @@ where
         self
     }
 
+    /// Allows setting specific jose types, the verifier will check against during verification.
```
ramosbugs, line 190: nit: JOSE

src/verification/mod.rs
```diff
@@ -203,14 +231,33 @@ where
...
+        // see https://tools.ietf.org/html/rfc7519#section-5.1
         if let Some(ref jwt_type) = jose_header.typ {
-            if jwt_type.to_uppercase() != "JWT" {
-                return Err(ClaimsVerificationError::Unsupported(format!(
-                    "unexpected or unsupported JWT type `{}`",
-                    **jwt_type
-                )));
+            if let Some(allowed_jose_types) = &self.allowed_jose_types {
+                // Check according to https://tools.ietf.org/html/rfc7515#section-4.1.9
+                // See https://tools.ietf.org/html/rfc2045#section-5.1 for a the full Content-Type Header Field spec.
```
ramosbugs, line 238: nit: `for a the` -> `for the`

src/verification/mod.rs
```diff
@@ -569,6 +616,18 @@ where
         self
     }
 
+    /// Specifies which JSON Web Token Header types are allowed
+    pub fn set_allowed_jose_types(mut self, types: Vec<JsonWebTokenType>) -> Self {
```
ramosbugs, line 620: let's use `IntoIterator` as with `set_allowed_algs`

ramosbugs (commented):
I've gone ahead and reverted the other PR, so this should apply cleanly against the latest commit now. just some minor comments and then I'd be happy to merge.

will118:
Apologies, I misunderstood the status of this PR. I thought it had stalled on the review comment. 

Looking forward to this being merged.

lmm-git:
I addresses your comments and fixed some more typos in ddf65a7f2eeabd658786e50f0e634e25d03952d1

With 03f021eda6c8c23a6d7e817073a5bfa0963b0b9e I introduced a new type of a normalized json web token type. I am really not sure whether it makes the code better, but some of my reasons are:

* De-couple the parsing of the type field from the verification
* It is clear whether some function handles a normalized type value or not without any runtime penalties
* Some can use the parsing not only during verification

However, I see some reasons why not to include this commit:

* It creates a new type, which type is uncommon for this crate
* It makes understanding the code a bit more complicated
* Users might create normalized objects containing non normalized values (I kept this possibility to allow be able to use the `new_type` macro and to allow creating these values from constants)

So what do you think? I am completely fine with just leaving out the second commit. It was just an experiment of me whether it makes the code better (and I am really not sure whether it does or not)

src/jwt/mod.rs
```diff
@@ -20,11 +20,98 @@ new_type![
     JsonWebTokenContentType(String)
 ];
 
+#[derive(Error, Debug)]
+pub enum JsonWebTokenTypeNormalizeError {
```
ramosbugs, line 24: let's make this `#[non_exhaustive]` to minimize future breaking changes

src/jwt/mod.rs
```diff
@@ -20,11 +20,98 @@ new_type![
...
+    type Error = JsonWebTokenTypeNormalizeError;
+
+    /// Normalizes a [`JsonWebTokenType`] and produces a [`NormalizedJsonWebTokenType`] according to RFC2045.
+    ///
+    /// See [RFC 2045 section 5.1](https://tools.ietf.org/html/rfc2045#section-5.1) for the full Content-Type Header Field spec.
+    /// See [RFC 7515 section 4.19](https://tools.ietf.org/html/rfc7515#section-4.1.9) for specific requirements of JSON Web Token Types.
+    fn try_from(t: &JsonWebTokenType) -> Result<NormalizedJsonWebTokenType, Self::Error> {
+        let lowercase_jwt_type = t.0.to_lowercase();
+        if lowercase_jwt_type.contains('/') {
+            if lowercase_jwt_type.starts_with("application/") {
```
ramosbugs, line 100: where does the spec require MIME types to be `application/*`? I see where that's implied when omitted, but I don't see why some custom app couldn't have its own complex MIME type with a different top-level type

src/jwt/mod.rs
```diff
@@ -20,11 +20,98 @@ new_type![
...
+        pub fn normalize(&self) -> Result<NormalizedJsonWebTokenType, JsonWebTokenTypeNormalizeError> {
+            self.try_into()
+        }
+    }
+];
+
+new_type![
+    /// JSON web token type field (typ)
+    ///
+    /// This types stores the raw (deserialized) value.
```
ramosbugs, line 57: this is the normalized value, not the raw one, right?

src/jwt/mod.rs
```diff
@@ -20,11 +20,98 @@ new_type![
...
+        /// See [RFC 2045 section 5.1](https://tools.ietf.org/html/rfc2045#section-5.1) for the full Content-Type Header Field spec.
+        /// See [RFC 7515 section 4.19](https://tools.ietf.org/html/rfc7515#section-4.1.9) for specific requirements of JSON Web Token Types.
+        pub fn normalize(&self) -> Result<NormalizedJsonWebTokenType, JsonWebTokenTypeNormalizeError> {
+            self.try_into()
+        }
+    }
+];
+
+new_type![
+    /// JSON web token type field (typ)
```
ramosbugs, line 55: `web token` -> `Web Token`

src/jwt/mod.rs
```diff
@@ -20,11 +20,98 @@ new_type![
...
+    ComplexTypeWithoutApplication(),
+}
+
 new_type![
+    /// JSON web token type field (typ)
+    ///
+    /// This types stores the raw (deserialized) value.
+    ///
+    /// To compare two different JSON Web Token types, please use the normalized version via [`JsonWebTokenType::normalize`]
     #[derive(Deserialize, Hash, Ord, PartialOrd, Serialize)]
```
ramosbugs, line 40: I'm noticing it's odd that `PartialOrd` and `Ord` are implemented but `PartialEq` and `Eq` aren't. If we don't want users to compare these non-normalized types to one another, then we should probably remove `PartialOrd`, `Ord`, and `Hash` too (not a breaking change since this type wasn't public before)

src/jwt/mod.rs
```diff
@@ -20,11 +20,98 @@ new_type![
...
+    #[error(
+        "The web token type is complex (contains /) but does not start with an application type."
+    )]
+    ComplexTypeWithoutApplication(),
+}
+
 new_type![
+    /// JSON web token type field (typ)
+    ///
+    /// This types stores the raw (deserialized) value.
```
ramosbugs, line 37: `types` -> `type`

src/jwt/mod.rs
```diff
@@ -20,11 +20,98 @@ new_type![
...
+    ParameterWithoutApplication(),
+
+    #[error(
+        "The web token type is complex (contains /) but does not start with an application type."
+    )]
+    ComplexTypeWithoutApplication(),
+}
+
 new_type![
+    /// JSON web token type field (typ)
```
ramosbugs, line 35: `web token` -> `Web Token`

src/jwt/mod.rs
```diff
@@ -20,11 +20,98 @@ new_type![
...
+    ///
+    /// This types stores the raw (deserialized) value.
+    ///
+    /// To compare two different JSON Web Token types, please use the normalized version via [`JsonWebTokenType::normalize`]
```
ramosbugs, line 39: (period at the end)

src/jwt/mod.rs
```diff
@@ -20,11 +20,98 @@ new_type![
...
+    /// See [RFC 2045 section 5.1](https://tools.ietf.org/html/rfc2045#section-5.1) for the full Content-Type Header Field spec.
+    /// See [RFC 7515 section 4.1.9](https://tools.ietf.org/html/rfc7515#section-4.1.9) for specific requirements of JSON Web Token Types.
+    ///
+    /// It is recommended to only instantiate `NormalizedJsonWebTokenType` objects via [`JsonWebTokenType`] and then call [`JsonWebTokenType::normalize`].
+    ///
+    /// ```rust
+    /// # use openidconnect::{NormalizedJsonWebTokenType, JsonWebTokenType};
+    /// let token_type: NormalizedJsonWebTokenType = JsonWebTokenType::new("jwt+at".to_string())
+    ///     .normalize()
+    ///     .unwrap();
```
ramosbugs, line 69: I'd rather not include `.unwrap()` in the documentation snippet since it encourages users not to properly handle errors. can we wrap it in a (hidden using `#`) function that returns a `Result` and then change this to `.normalize()?`?

src/jwt/mod.rs
```diff
@@ -20,11 +20,98 @@ new_type![
...
+    ///
+    /// // normalized value looks like this
+    /// assert_eq!(*token_type, "application/jwt+at")
+    /// ```
+    ///
+    /// When you really want to create this type directly, please check the following items:
```
ramosbugs, line 75: this really suggests that we shouldn't allow direct construction (without `normalize` or `TryFrom`), which means not using `new_type!`. it shouldn't be too bad to implement this type manually though (without the `new` function): https://github.com/ramosbugs/openidconnect-rs/blob/fd404985ef6c8e546f951191f4e1bc791615f5ca/src/macros.rs#L122-L146

ramosbugs (commented):
Types are a great way to capture this sort of thing, so I like this approach!

src/jwt/mod.rs
lmm-git: good catch!

I changed the logic to match the standard now (hopefully ^^).

Also added some tests.

src/jwt/mod.rs
lmm-git: I think they are implemented through https://github.com/ramosbugs/openidconnect-rs/blob/02525323bfdcef1c0da17835fa0d9738db354765/src/macros.rs#L161

But I removed `PartialOrd` and `Ord`.

Generally I also think that comparing these values shouldn't be (easily) possible as this might lead to wrong results. The more interesting derive would therefore be `Eq`, but as far as I can see the macro does not support switching off that specifically.

src/jwt/mod.rs
lmm-git: There is one more "web token" in this file. Should I fix this too (maybe in an extra commit)?

src/jwt/mod.rs
lmm-git: I am actually not sure what you mean. I removed the `unwrap` from shown documentation at least.

src/jwt/mod.rs
```diff
@@ -20,11 +20,113 @@ new_type![
     JsonWebTokenContentType(String)
 ];
 
+/// Error type used when normalizing [`JsonWebTokenType`] objects
+#[derive(Error, Debug)]
+#[non_exhaustive]
+#[allow(clippy::enum_variant_names)] // new error types might not start with InvalidMime
```
ramosbugs, line 26: now that this is up to 3 variants that are very similar, I think it may be preferable to have a single error type that doesn't distinguish between the different reasons for the MIME type to be invalid. e.g.,
```rust
#[derive(Error, Debug)]
#[error("Invalid JWT type: {typ}")]
pub struct InvalidJsonWebTokenTypeError {
  typ: String,
}
```
this also includes the offending type, which should help users debug the issue. I'm assuming it'll be pretty obvious to users why the printed MIME type isn't valid 🤷‍♂️

src/verification/mod.rs
```diff
@@ -140,6 +141,15 @@ where
                     .cloned()
                     .collect(),
             ),
+            allowed_jose_types: Some(HashSet::from([
+                JsonWebTokenType::new("application/jwt".to_string())
+                    .normalize()
+                    .unwrap(), // used by many IdP, but not standardized
```
ramosbugs, line 147: since this is the production code path, let's use `.expect("application/jwt should be a valid JWT type")` instead of `.unwrap()` so that the panic message is at least somewhat informative if our assumptions are wrong

src/verification/mod.rs
```diff
@@ -140,6 +141,15 @@ where
...
+                JsonWebTokenType::new("application/jose".to_string())
+                    .normalize()
+                    .unwrap(), // standard as defined in https://tools.ietf.org/html/rfc7515#section-4.1.9
```
ramosbugs, line 150: `.expect(...)`

src/verification/tests.rs
```diff
@@ -47,16 +100,125 @@ fn test_jose_header() {
...
+
+        custom_verifier
+            .validate_jose_header(
+                &serde_json::from_str::<CoreJsonWebTokenHeader>(
+                    "{\"alg\":\"RS256\",\"typ\":\"NOT_A_JWT_REALLY\"}",
+                )
+                .expect("failed to deserialize"),
+            )
+            .expect("JWT type should be allowed but is not");
+    }
```
ramosbugs, line 210: let's add one more case with a parameterized MIME type (i.e., with a `;`) since those should be accepted here

ramosbugs (commented):
thanks for addressing the comments! almost ready to merge :-)

src/jwt/mod.rs
lmm-git: I thought it might be nice to allow error differentiation here but I see that it is much more simple without it.

src/verification/tests.rs
ramosbugs: have you had a chance to add this test case?

lmm-git: Ah sorry, I did not read exactly and thought you meant another case. Now a test case should be added.

ramosbugs:
Thanks!

lmm-git:
Thank you for your extensive reviews!

PR #175, opened 2024-06-10
Adds `IdTokenVerifier::require_typ_check` method

will118:
This crate validates the `typ` field of the JOSE header equals `"JWT"` case-insensitively.

However, some identity providers return tokens with `typ` values such `"at+jwt"` which are largely compatible but impossible to use with this crate.

This change introduces a builder method on `IdTokenVerifier` which allows disabling this check for users that wish to.

The current behavior is retained as the default (i.e. it requires the value to equal `"JWT"`).

### DIFF
--- a/src/verification/mod.rs
+++ b/src/verification/mod.rs
@@ -125,6 +125,7 @@ where
     iss_required: bool,
     issuer: IssuerUrl,
     is_signature_check_enabled: bool,
+    is_typ_check_enabled: bool,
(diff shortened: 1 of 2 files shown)


will118:
There is already a PR in-flight that does something similar: #161 but hasn't seen much activity in the last couple months, but did save me a lot of time in making this change.

I ran into this today with tokens that have  `typ="at+jwt"` and figured I would open a quick PR in the hope it can be merged.

ramosbugs:
Thank you!

Issue #153, opened 2024-03-27
Add examples for non-oidc-provider providers mentioned in README


Nerglej:
The README for this repo has this line as the very first you read:

```
This library provides extensible, strongly-typed interfaces for the OpenID Connect protocol, which can be used to authenticate users via GitHub, Google, GitLab, Microsoft, Apple, Discord, and many other providers.
```

There's only an example in the docs for [Google](https://github.com/ramosbugs/openidconnect-rs/blob/main/examples/google.rs) which uses Google's discover endpoint (/.well-known/openid-configuration).

The README clearly mentions GitHub, Discord and 'many other providers', but GitHub and Discord (the ones that I've tested, but I would imagine a lot more) doesn't have a discover endpoint.

That is in itself fine, because you can just work around the discover-endpoint with `CoreProviderMetadata`. You can't use `CoreProviderMetadata::new()` for Discord though, since they also don't have a JWKS-endpoint (haven't checked GitHub). And that's where I'm currently stuck, but will still work on it though.

So I'm kindly asking for an example of how to implement an OpenID Connect client for Discord/GitHub/'many other providers' that doesn't have the discover endpoint. I've already implemented authentication for Discord, GitHub, Facebook and others with the `OAuth2` crate, so I'm not that new to this, but I still feel like I'm missing something in this crate...

Kind regards😊

ramosbugs:
> GitHub, Google, GitLab, Microsoft, Apple, Discord, and many other providers.

Ah whoops, I think I was confusing GitHub's support for OIDC [from within Actions](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-cloud-providers) for its own API auth, which seems to be OAuth2-only. I'll remove GitHub from that list. If I was mistaken about any other providers in this list supporting OIDC, I'd be happy to remove those as well.

> So I'm kindly asking for an example of how to implement an OpenID Connect client for Discord/GitHub/'many other providers' that doesn't have the discover endpoint. I've already implemented authentication for Discord, GitHub, Facebook and others with the `OAuth2` crate, so I'm not that new to this, but I still feel like I'm missing something in this crate...

This is fair, and I agree it's worth adding an example that shows how to set the endpoints manually.

[...]

PR #181, opened 2024-07-15
Add Discord example

yasamoka:
Related to #153.

Thanks!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,10 @@ timing-resistant-secret-traits = ["oauth2/timing-resistant-secret-traits"]
 ureq = ["oauth2/ureq"]
 
 [[example]]
+name = "discord"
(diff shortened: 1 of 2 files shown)


inklesspen:
Did you test this? If it worked at one point, it doesn't work now; Discord doesn't return an `id_token`.

See https://discord.com/developers/docs/topics/oauth2#authorization-code-grant-access-token-exchange-example for what they do return.

yasamoka:
> Did you test this? If it worked at one point, it doesn't work now; Discord doesn't return an `id_token`.
> 
> 
> 
> See https://discord.com/developers/docs/topics/oauth2#authorization-code-grant-access-token-exchange-example for what they do return.

I believe I did. This is 1.5 years old at this point and probably some things have changed.

Would welcome a superseding PR if there is any benefit to it.

Issue #122, opened 2023-07-24
Strict validation breaks due to Microsoft's OIDC noncompliance


Redrield:
This library strictly validates responses from the server according to spec in ways that make it incompatible with Microsoft's implementation of OIDC in certain situations [due to their refusal to make their implementation spec compliant](https://github.com/MicrosoftDocs/azure-docs/issues/38427#issuecomment-801417083). Currently the library doesn't allow users to configure the validation strictness, meaning the only way to work around this is to copy the repository locally to selectively ignore validation failures caused by this. It would be helpful if users could configure this to disable the validation to support Microsoft OIDC (Or alternatively, to specify alternate valid fields to provide some level of certainty that the payload is mostly correct).

The two checks that fail are in verification.rs:284 (JWT issuer validation), and discovery.rs:386 (Provider metadata issuer validation).

Based on the responses to the above issue (as well as the age of the issue), I don't think a spec compliant implementation of OIDC should be expected from Microsoft any time soon, and as such it would be helpful to allow for exceptions in this case to be compatible.

ramosbugs:
I'm sympathetic to Microsoft's customers' frustrations here while also being annoyed that Microsoft decided to ignore the spec and push this problem to both customers and library maintainers. Since this library focuses on *standardized* protocols, I think a reasonable goal here is to have escape hatches (but not foot guns) available that allow users to interoperate with Microsoft, but without adding any Microsoft-specific functionality to this crate. A PR adding a working Microsoft code example would be welcomed, though.

This means that Microsoft customers should be able to use this crate, but their code may be more verbose than for users of standards-compliant OIDC implementations. Making Microsoft OIDC integrations work out of the box without any extra code is a non-goal of this crate, although community members should feel free to publish a separate crate built on top of this one to improve the ergonomics.

> The two checks that fail are in verification.rs:284 (JWT issuer validation), and discovery.rs:386 (Provider metadata issuer validation).

For JWT issuer validation, there are already two escape hatches available:
1. When [instantiating an `IdTokenVerifier`](https://docs.rs/openidconnect/3.3.0/openidconnect/struct.IdTokenVerifier.html#method.new_public_client), an `IssuerUrl` should be specified that matches the one that will be present in the ID tokens. This is the recommended approach.
[...]

PR #183, opened 2024-09-15
Add ProviderMetadata::discover_with_options and "common" Entra example

phy1729:
#179 handles the case for an Entra application that allows signins from a single tenant. For Entra applications that allow signins from personal and enterprise accounts, the `https://login.microsoftonline.com/common/v2.0` issuer is used which returns a metadata document with `issuer` set to `https://login.microsoftonline.com/{tenantid}/v2.0`.

This PR adds `ProviderMetadata::discover_with_options` and `ProviderMetadata::discover_async_with_options` to disable the `issuer` validation required by the OIDC spec. I think this is more discoverable than the proposed route in #122 of wrapping http_client to munge the response.

In the example `require_issuer_match` is disabled on the `IdTokenVerifier` because the ID token has the `iss` claim set with the specific tenant ID of the account which is not a static value.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,10 @@ timing-resistant-secret-traits = ["oauth2/timing-resistant-secret-traits"]
 ureq = ["oauth2/ureq"]
 
 [[example]]
+name = "entra"
+required-features = ["reqwest-blocking"]
+
+[[example]]
 name = "gitlab"
 required-features = ["reqwest-blocking"]
 
--- a/examples/entra.rs
+++ b/examples/entra.rs
@@ -1,0 +1,185 @@
+//!
+//! This example showcases the process of integrating with the
+//! [Microsoft Entra OpenID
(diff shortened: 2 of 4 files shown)


maxammann:
thank you! @ramosbugs this seems required for typical SSO setups with Ms

maxammann:
A workaround is to do:

```
        let client = CoreClient::new(
            ClientId::new(client_id.to_string()),
            issuer_url,
            JsonWebKeySet::fetch_async(
                &JsonWebKeySetUrl::new("https://login.microsoftonline.com/common/discovery/v2.0/keys".to_string())?,
                &http_client,
            )
            .await?,
        )
        .set_auth_uri(oauth2::AuthUrl::new(
            "https://login.microsoftonline.com/common/oauth2/v2.0/authorize".to_string(),
        )?)
        .set_token_uri(oauth2::TokenUrl::new(
            "https://login.microsoftonline.com/common/oauth2/v2.0/token".to_string(),
        )?)
```

PR #207, opened 2025-05-15
Allow for optional introspection URL

weichweich:
Allows to use the Client with an optional IntrospectionURL.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -494,6 +494,46 @@ where
         }
     }
 
+    /// Set the [RFC 7662](https://tools.ietf.org/html/rfc7662) introspection endpoint.
+    ///
+    /// See [`introspect()`](Self::introspect).
+    pub fn set_introspection_url_option(
+        self,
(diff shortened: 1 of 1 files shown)


silwol:
Any chance to get this merged, @ramosbugs ? I'd love to upgrade from openidconnect v3 to v4, this change would be the enabler for that.

aitiotekt:
Perhaps it would be best to add all of them (excluding `set_auth_uri_option`, which is required in the openidconnect scenario): 

- `set_introspection_url_option` 
- `set_revocation_url_option` 
- `set_device_authorization_url_option`
- `set_token_uri_option` 
- `set_user_info_url_option`

Since the `oauth2` crate supports the first four, and it would also be best to add the last one for symmetry.

[CAN FIND THEM HERE](https://github.com/ramosbugs/openidconnect-rs/compare/main...ethaxon:openidconnect-rs:set_option)

PR #211, opened 2025-07-03
fix: return valid audience when multiple are provided

### LINKED ISSUE #210: Invalid audience
I've built an app that uses openidconnect-rs that lets folks bring their own OIDC providers. I've developed against Auth0 and Discord, however, when using Zitadel and Frontegg, I'm unable to authenticate and get an error 

> Invalid audiences: must contain `my-client-id` (found audiences: `8d3b3f49--id-that-looks-similar-to-client-id`)

I think this may be related to the jwk? For the frontegg OIDC provider the invalid audience _always_ has the prefix from the `kid` field from the [jwks_uri](https://app-mxi106np7cv9.frontegg.com/.well-known/jwks.json) link. Similarly, from the Zitadel side, the audience is always one of the `kid` from the [jwks_uri](https://ricochet-test-h7fb4u.us1.zitadel.cloud/oauth/v2/keys) from the .well-known/openid-configuration. 

I am following the GitLab example to a T for my OIDC auth configuration. 

Is this an issue with my provider discover or other?

JosiahParry:
This PR  closes #210

### DIFF
--- a/src/verification/mod.rs
+++ b/src/verification/mod.rs
@@ -343,7 +343,6 @@ where
                         .iter()
                         .any(|aud| (**aud).deref() == self.client_id.deref());
 
-                    log::debug!("Audience found in audiences {audience_matched}");
                     if !audience_matched {
                         return Err(ClaimsVerificationError::InvalidAudience(format!(
                             "must contain `{}` (found audiences: {})",


src/verification/mod.rs
```diff
@@ -353,17 +355,25 @@ where
...
-                            .filter(|aud| (**aud).deref() != self.client_id.deref())
-                            .find(|aud| !(self.other_aud_verifier_fn)(aud))
-                            .map(|aud| {
-                                Err(ClaimsVerificationError::InvalidAudience(format!(
-                                    "`{}` is not a trusted audience",
-                                    **aud,
-                                )))
-                            })
-                            .unwrap_or(Ok(()))?;
+                            .find(|aud| aud.as_str().eq(self.client_id.as_str()));
```
JosiahParry, line 360: First check that the client id is in the provided audience. Only when it isn't found do we apply the other audience verifier.

PR #212, opened 2025-07-03
fix: return valid audience when multiple are provided 

### LINKED ISSUE #210: Invalid audience

JosiahParry:
This PR closes #210 by first checking if the client ID is in the provided audiences. Otherwise perform the verifier function. 

This is a stylistically different approach than in #211 which I wrote to help debug the problem first.

### DIFF
--- a/src/verification/mod.rs
+++ b/src/verification/mod.rs
@@ -353,23 +353,33 @@ where
                     }
 
                     if let None = known_audience {
-                        let e_msg = if audiences.len() > 1 {
-                            let provided_audiences = audiences
-                                .iter()
-                                .map(|aud| format!("`{}`", Deref::deref(aud)))
-                                .collect::<Vec<_>>()
-                                .join(", ");
-                            format!("must contain `{expected_audience}` (found audiences {provided_audiences})")
-                        } else {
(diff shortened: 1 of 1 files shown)


brianbento:
Waiting for this to get merged in so that a lot of rust projects will start working with Zitadel again. 

Thanks @JosiahParry!

